PR #11, opened 2014-07-30
FIX Continue Header

iannsp:
add '\r\n' to the header
update test expectation
close reactphp/react#308


slava-vishnyakov:
This seems to have been planned to release in 0.4.1, but that tag seems to be deleted?

If anyone stumbles into this - as of `0.4` you need to use `dev-master` explicitly

```        "react/http": "dev-master",
```

PR #12, opened 2014-08-25
Answer with HTTP/1.0 until 1.1 is really implemented

nicolas-grekas:
Currently, the HTTP server answers with `1.1` but does not handle persistent connexions.
Because of this, well behaving clients, expecting proper handling of the advertised 1.1 semantic do not work:
the incoming Request object does never trigger the `end` event.
This of course is because an HTTP/1.1 client never closes the socket connexion.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -12,7 +12,8 @@ class Response extends EventEmitter implements WritableStreamInterface
     private $writable = true;
     private $conn;
     private $headWritten = false;
-    private $chunkedEncoding = true;
+    // TODO chunked is a HTTP/1.1 feature that needs proper Request parsing before enabling
+    private $chunkedEncoding = false;
 
     public function __construct(ConnectionInterface $conn)
     {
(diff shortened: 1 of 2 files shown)


clue:
What is the status of this PR?

Does this really fix any bugs? If so, can we add some test cases to confirm this?

FWIW, chunked transfer encoding is currently being implemented by @WyriHaximus here: https://github.com/reactphp/http-client/pull/58

clue:
Thanks for filing this PR @nicolas-grekas! :+1: 

I think this addresses a real issue, but I'd rather address the `end` event via #104 and chunked transfer encoding via #96.

This is kind of an old PR and it contains plenty of merge conflicts now. Unfortunately, this hasn't been updated in years, so I'll close this due to a lack of feedback. If you think this is still worth addressing, please update this PR or file a new one and we can reconsider this :+1:

clue:
For the reference: Persistent connections will be addressed via #39 (eventually), but I'll look into explicitly denying persistent connection support for HTTP/1.1 clients in the meantime :+1:

clue:
The issues you were seeing should now be addressed via #128.

I hope this helps :+1:

nicolas-grekas:
not sure about that: the *client* sends chunked encoding in my case.

clue:
Incoming chunked transfer encoding will now be decoded via #116 :shipit:

PR #13, opened 2014-10-09
Multipart handling

onigoetz:
Hello,

I added basic support for multipart for files and POST requests

I also wrote the unit tests for this part and improved the existing ones.

One thing that could be done better is that now, it waits for the whole body before it parses it, we could implement a system where it feeds the content to the parser and the files are created on the fly and we don't use the whole memory for file uploads.

### DIFF
--- a/src/MultipartParser.php
+++ b/src/MultipartParser.php
@@ -142,16 +142,26 @@ class MultipartParser
         $content = preg_replace('/Content-Type: (.*)[^\n\r]/', '', $match[3]);
         $content = ltrim($content, "\r\n");
 
-        $path = tempnam(sys_get_temp_dir(), "php");
-        $err = file_put_contents($path, $content);
+        // Put content in a stream
+        $stream = fopen('php://memory', 'r+');
+        if ($content !== '') {
+            fwrite($stream, $content);
+            fseek($stream, 0);
+        }
 
         $data = [
             'name' => $match[2],
             'type' => trim($mime[1]),
-            'tmp_name' => $path,
-            'error' => ($err === false) ? UPLOAD_ERR_NO_FILE : UPLOAD_ERR_OK,
-            'size' => filesize($path),
+            'stream' => $stream, // Instead of writing to a file, we write to a stream.
+            'error' => UPLOAD_ERR_OK,
+            'size' => function_exists('mb_strlen')? mb_strlen($content, '8bit') : strlen($content),
         ];
+
+        //TODO :: have an option to write to files to emulate the same functionality as a real php server
+        //$path = tempnam(sys_get_temp_dir(), "php");
+        //$err = file_put_contents($path, $content);
+        //$data['tmp_name'] = $path;
+        //$data['error'] = ($err === false) ? UPLOAD_ERR_NO_FILE : UPLOAD_ERR_OK;
 
         $this->addResolved('files', $match[1], $data);
     }
--- a/src/MultipartParser.php
+++ b/src/MultipartParser.php
@@ -1,0 +1,193 @@
+<?php
+
+namespace React\Http;
+
+/**
+ * Parse a multipart body
+ *
+ * Original source is from https://gist.github.com/jas-/5c3fdc26fedd11cb9fb5
+ *
+ * @author jason.gerfen@gmail.com
+ * @author stephane.goetz@onigoetz.ch
+ * @license http://www.gnu.org/licenses/gpl.html GPL License 3
+ */
+class MultipartParser
+{
+    /**
+     * @var string
+     */
+    protected $input;
+
+    /**
+     * @var string
+     */
+    protected $boundary;
+
+    /**
+     * Contains the resolved posts
+     *
+     * @var array
+     */
+    protected $post = [];
+
+    /**
+     * Contains the resolved files
+     *
+     * @var array
+     */
+    protected $files = [];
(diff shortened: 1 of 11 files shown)


src/MultipartParser.php
```diff
@@ -0,0 +1,193 @@
...
+    protected function file($string)
+    {
+        preg_match('/name=\"([^\"]*)\"; filename=\"([^\"]*)\"[\n|\r]+([^\n\r].*)?\r$/s', $string, $match);
+        preg_match('/Content-Type: (.*)?/', $match[3], $mime);
+
+        $content = preg_replace('/Content-Type: (.*)[^\n\r]/', '', $match[3]);
+        $content = ltrim($content, "\r\n");
+
+        $path = tempnam(sys_get_temp_dir(), "php");
+        $err = file_put_contents($path, $content);
```
WyriHaximus: Imho it would be better when you use streams and pass those along. That way the developer can decide what to do with the contents of the stream.

onigoetz: You think one stream per fileupload or the raw stream ?

I'll look how I can do that, I'm not a stream expert

WyriHaximus: A stream per fileupload. Ping me on IRC or on twitter if you have questions about that. Happy to help :).

nazar-pc: I have suggestion about how to do that:
* add one more argument (callable or factory) for handling file uploads to server constructor
* then when it comes to file handling we just call callable or factory (with arguments like file name, mime type) and pass everything to method of object it returned (if there was no callable or factory - just ignore uploaded file, it is too expensive to store it in memory)
* after whole request parsed we'll have access to array with all objects we've created for uploaded files

What you think about such approach?

onigoetz: Thanks for your ideas, I will try to make it this week.

WyriHaximus:
@onigoetz after reading you're message here https://github.com/reactphp/react/issues/3#issuecomment-78455082 want me to look into that for you and make a PR on your PR?

onigoetz:
@WyriHaximus yes I would be very happy if you can help on the matter.

WyriHaximus:
@onigoetz I'll look into it tonight or tomorrow :+1:

HonzaMac:
Any progress on that?

onigoetz:
None one my part, @WyriHaximus were you able to take a look ?

WyriHaximus:
I did take a look, started writing some code and got stuck at the time.
Will revisit it soon to see if I can work it out this time.

onigoetz:
Hi guys, I updated my fork to be up to date with all the changes on the master branch.

I also changed the logic to not store the file upload in temp files, but keep them in a stream.

I think this PR could be merged as-is and another pull-request could finish the work I started. What do you think ?

WyriHaximus:
@onigoetz cheers looks good, will have a go with it again this weekend. My apologies for not getting back to you last time. I've looked into converting it into a stream streaming files and that looked harder then expected at the time. Going to either a special stream/buffer for it and do a follow up PR getting that in.

This next bit is a braindump from a previous chat on IRC about it. The idea was to emit events on files found in the stream and also on multipart bits.

Since steaming support also has to be done I can include the above bit in that PR as well :).

WyriHaximus:
I'll follow this up shortly with another PR before tagging it. Thanks for the work :+1:

nazar-pc:
Can't believe it was finally merged) Yay!

onigoetz:
:+1: awesome !

WyriHaximus:
@onigoetz will try to dedicate as mush OSS time as I can on the follow up. Will comment here when the initial WIP PR is up

HonzaMac:
Awesome! Cant believe tooo :+1:

clue:
As much as I'd love to see multipart handling in here, I have to admit I'm not a big fan of the current implementation.

I think it's important to consider that HTTP request bodies can be of arbitrary sizes. While a common form submission only results in a few (kilo)bytes of data, it's quite common to use HTTP-based APIs to transmit multi-megabytes of data. The same applies to uploading files via HTML forms or using CLI-based tools to trigger similar HTTP requests.

Admittedly, a streaming approach where only small chunks are kept in memory is probably slightly more complicated to implement and likely more difficult to consume from a user perspective.

FWIW, a streaming approach could probably look somewhat similar to a TAR-file decoder which emits an event for each *part* and then only streams its *contents*. I have to admit I hate posting links to my own libraries, but perhaps https://github.com/clue/php-tar-react could be a good start here?

Any considerations?

WyriHaximus:
@clue working on a streaming implementation inspired by clue/php-tar-react :+1:

clue:
@WyriHaximus Awesome, keep me posted! Would love to take a look at this eventually

WyriHaximus:
@clue will do, gained some good experience with out-of-the-box streams in https://github.com/reactphp/filesystem so it should be a matter get getting into the flow

onigoetz:
I admit my first implementation was very naive and I wanted to make a streaming implementation after a first working version, sadly I did not have time to do so.

"""



---
Stéphane Goetz
Développeur Web

http://www.onigoetz.ch

Mobile : +41 79/824.77.68
Mail : stephane.goetz@onigoetz.ch

WyriHaximus:
#41 is the follow up PR, feedback is welcome

PR #21, opened 2015-02-27
update guzzle dependency

tomdmaguire:
guzzle/parser is no longer available


cboden:
Unfortunately guzzle/guzzle introduces an un-needed dependency restriction in ext-curl. While this PR will fix the composer warning it creates a BC break.

cboden:
resolved by #29

PR #23, opened 2015-03-07
changed "guzzle/parser" to "guzzle/guzzle"

cystbear:
as far as 
```
This package is abandoned and no longer maintained. The author suggests using the guzzle/guzzle package instead.
```


tomdmaguire:
dupe of https://github.com/reactphp/http/pull/21

PR #33, opened 2015-08-12
Interpret content-type header for 'application/x-www-form-urlencoded' correctly

caseyamcl:
Sometimes the `Content-Type` header can have additional information, such as `charset`.  Also, there should only be one Content-Type header in a HTTP Request.

This PR fixes interpretation of the `application/x-www-form-urlencoded` by comparing only the beginning of the `Content-Type` header value instead of testing for equality.  So, now both of these headers will now be interpreted correctly:

Worked before:

    Content-Type: application/x-www-form-urlencoded

Works now:

    Content-Type: application/x-www-form-urlencoded; charset=utf-8

### DIFF
--- a/.idea/composerJson.xml
+++ b/.idea/composerJson.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ComposerJsonPluginSettings">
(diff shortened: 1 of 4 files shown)


caseyamcl:
Closing (accidentally added `.idea` crap to the commit

PR #46, opened 2015-11-25
Support json content in RequestParsers class

Abdulrazak-Alkl:
This PR to Support new HTTP content-type "application/json".
Currently body property in request object will contains a JSON object decoded from incoming JSON string.

### DIFF
--- a/src/RequestParser.php
+++ b/src/RequestParser.php
@@ -140,6 +140,12 @@ class RequestParser extends EventEmitter
 
                 return;
             }
+
+            if (strtolower($headers['Content-Type']) == 'application/json') {
+                $result = json_decode($content);
+                $this->request->setBody($result);
+                return;
+            }
         }
 
 


ephrin:
-1 I believe mixing in framework like "helpers" to server things is not a good idea.

marcj:
+1, this library's purpose is to marshal data from tcp to PHP objects. When a client sends a file, it converts it to a FileInfo object. When urlencoded, it converts it to php array. So, when the client sends json, this library should also convert this into a php array. The only conclusion you could make. Actually I as a React\Request object consumer don't care which content-type the client used. I want to sent data, not in a raw format.

staabm:
Not sure if FileInfo is appropriate because it might require to load a file into memory which wouldnt work for big/huge files. Something streamable would be good.

clue:
Afaict this has been filed to build a new feature on top of the previously broken master branch. Now that #59 is in, is this still relevant?

clue:

PR #49, opened 2016-01-12
Multipart multiplying memory allocation

pigi72333:
When we parse file by preg_match we duplicate memory allocation for it. 
For example: when uploading 10M after preg_match, we have ~30M memory allocated, 10M from request body and ~20M in matches array, because matches will contain entry matches of pattern and masks.


clue:

pigi72333:
not actual, now

clue:

PR #53, opened 2016-03-22
Add new HeaderBag (modeled after Symfony) to support case-insensitive header matching.

pwhelan:
Used a new HeaderBag class to support case-insensitive header name matches. This fixes a bug where ab (Apache Benchmark) simply hangs.

### DIFF
--- a/src/HeaderBag.php
+++ b/src/HeaderBag.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace React\Http;
+
+
+class HeaderBag extends \ArrayObject
+{
+    /** Map Header keys in a case-insensitive manner.
+     * This is a map from an upper case version of a header name
+     * to it's original name (so we can preserve case for posterity).
+     *
+     * @var array
+     */
+    protected $headermap = [];
+    
+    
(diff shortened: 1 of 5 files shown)


clue:
> support case-insensitive header name matches

Thanks for your effor @pwhelan! I agree that this a bug that needs fixing :+1: 

However, the current implementation contains a BC break and I see little we could do about this on this level. As an alternative, I would vote for waiting for #41 to remove most (all?) of the assumptions on headers.

FWIW: Proper header parsing will probably be resolved once we look into PSR-7 support (#28).

clue:

pwhelan:
> Afaict this has been filed to fix the previously broken master branch. Now that #59 is in, is this still relevant?

I'll have to test that to be sure. This bug was kept open pending the merge of #41 (at least on my end). 

I'll test this soon to be sure. This was meant primarily to fix problems with ab (Apache Benchmark) and other http clients (curl, possibly jmeter as well) which use 'Content-length' instead of 'Content-Length'. This happens enough in the wild that it has to be supported.

If #59 does fix this then it can be closed.

clue:

See also #51, which I'm currently looking into :+1:

PR #54, opened 2016-04-11
Add parameters to suppress sending X-Powered-By header

vfsoraki:
It is related to issue #36

### DIFF
--- a/src/ParamBag.php
+++ b/src/ParamBag.php
@@ -1,0 +1,39 @@
+<?php
+namespace React\Http;
+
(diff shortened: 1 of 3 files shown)


clue:
Thanks for your PR, but as much as I link this feature, I'm currently having to vote :-1: on this purely due to how much new code this introduces and how many tests this would require.

As an alternative, I would suggest we tackle this as part of #28, where we can likely introduce some kind of default response which would default to having this header applied.

PR #76, opened 2016-10-26
Handle keepalive

### LINKED ISSUE #39: Support persistent connections (aka HTTP/1.1 Keep-Alive)

jason69:
Hi,
I see a previous keepalive pull request opened for some time now.
Here is another one I made.
I find it clearer than the previous one because the connection is maintained/closed according to request header and a max number of reuse of the connection.

I've done some tests with apache benchark on a virtual machine and I'm getting an average of 1.6x requests/s with keepalive.

By the way testing with PHP 7.0.8 shows that the connexion shouldn't be reused too many times. I set it as a constant in the code to 25. Increasing this number shows a drop in speed, I havent managed to understand why...

Closes #39

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -13,6 +13,7 @@ class Response extends EventEmitter implements WritableStreamInterface
     private $conn;
     private $headWritten = false;
     private $chunkedEncoding = true;
+    public $closeConnection = false;
 
     public function __construct(ConnectionInterface $conn)
     {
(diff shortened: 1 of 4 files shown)


clue:
As much as I'd love to get this feature in, I'm having to close this for now as it hasn't received any input in a while and it's unlikely this will get traction any time soon.

The feature request is still open in #39 and we'll look into this in the not too far future :+1: If you feel this was closed prematurely or want to pick this up again, please let us know and we can reopen this.

Thank you for your effort!

PR #85, opened 2016-11-29
Fix pause/resume on request not working

### LINKED ISSUE #84: Pause/resume on request not working
Hi,

Why do we have this code in Server class?
````
$request->on('pause', function () use ($conn) {
    $conn->emit('pause');
});
$request->on('resume', function () use ($conn) {
    $conn->emit('resume');
});
````

I have changed emit's to correct methods (pause/resume), like this:
````
$request->on('pause', function () use ($conn) {
    $conn->pause();
});
$request->on('resume', function () use ($conn) {
    $conn->resume();
});
````

And now pausing works.

nick4fake:
Fixes #84


clue:
Ping @nick4fake, any updates here?

PR #89, opened 2017-02-05
update for phpunit 6

nopolabs:
Updated tests to run with phpunit/phpunit 6.0.*

### DIFF
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -14,7 +14,7 @@ class ResponseTest extends TestCase
         $expected .= "Transfer-Encoding: chunked\r\n";
         $expected .= "\r\n";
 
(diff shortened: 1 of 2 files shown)


jsor:
Thanks for the PR! As you can see, the tests on travis are failing because PHPUnit 6 is for PHP >= 7.0 only and react/http is >=5.4.0.

As per reactphp/react#354, the decision has been made to add phpunit/phpunit:^4.8 to require-dev for all components. See also https://github.com/reactphp/event-loop/pull/68#issuecomment-271815329 for how to avoid deprecation warnings for `getMock()`.

nopolabs:
thank you for pointing out the related issues

jsor:
@nopolabs Yes, sorry for the non-transparency. Care to file another PR? Otherwise i'll do one.

PR #115, opened 2017-02-11
Fix minimum phpunit version and allow v5

andig:
Follow-up of #111 and https://github.com/reactphp/http/pull/114

phpunit ^5.0 is required since as of 4.8.20 php 7.0 is not supported (see https://github.com/sebastianbergmann/phpunit/blob/4.8.32/ChangeLog-4.8.md)

Welcome to version hell :/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,6 @@
         }
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8"
+        "phpunit/phpunit": "^4.8.10||^5.0"
     }
 }
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -34,7 +34,9 @@ class ResponseTest extends TestCase
         $expected .= "Transfer-Encoding: chunked\r\n";
         $expected .= "\r\n";
 
-        $conn = $this->getMock('React\Socket\ConnectionInterface');
+        $conn = $this
+            ->getMockBuilder('React\Socket\ConnectionInterface')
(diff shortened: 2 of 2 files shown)


clue:
> phpunit ^5.0 is required since as of 4.8.20 php 7.0 is not supported (see https://github.com/sebastianbergmann/phpunit/blob/4.8.32/ChangeLog-4.8.md)

This change was actually reverted in 4.8.21. In other words: This actually works perfectly fine currently and installs 4.8.35 on PHP 7 (at the time of writing this).

That being said, I agree that it makes perfect sense to support PHPUnit 5 as well :+1:

composer.json
```diff
@@ -16,6 +16,6 @@
         }
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8"
+        "phpunit/phpunit": "^4.8.10|^5.0"
```
clue: Should probably be `^4.8.10 || ^5.0`?

Also, PHPUnit 5 emits several warnings on PHP 7, can you address these here? See also #93 and #107.

clue: Arguably, fixing the lowest version and supporting PHPUnit 5 *could* be two separate PRs, but I'll leave this up to you :+1:

andig:
@clue was a bit lazy and didn't split. Both phpunit 4 and 5 tests are passing now.

andig:
Once this is merged, https://github.com/reactphp/http/pull/114 should pass, too.

PR #122, opened 2017-02-17
The Server should always have a `request` listener

clue:
Having an HTTP server with no `request` listener does not really make any sense at all.

Currently, the `end()` call actually results in an invalid HTTP response stream, because it sends a closing chunk (chunked transfer encoding) without actually sending any HTTP response headers.

This simple PR changes it so that this case will simply not be handled at all, so that the client never receives any response data at all. This is actually in line with how our Socket component works if there's no `connection` listener.

Note that the invalid response stream will that is sent by `end()` will be addressed in a separate ticket, I'll link to this here once it's ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ $http->on('request', function (Request $request, Response $response) {
 
 See also [`Request`](#request) and [`Response`](#response) for more details.
 
+> Note that you SHOULD always listen for the `request` event.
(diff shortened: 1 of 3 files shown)


clue:

clue:
> Note that the invalid response stream will that is sent by end() will be addressed in a separate ticket, I'll link to this here once it's ready.

See #131

PR #126, opened 2017-02-19
Send HTTP status code 431 if request header is too large

clue:
Currently, sending an invalid request results in the same error message as sending a valid request that is too large: `HTTP/1.1 400 Bad Request`.

This simple PR change it so that the latter will now use the more descriptive error message: `HTTP/1.1 431 Request Header Fields Too Large`

Builds on top of #124

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -105,7 +105,7 @@ class Server extends EventEmitter
 
             $that->writeError(
                 $conn,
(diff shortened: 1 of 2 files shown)


clue:

clue:
Rebased now that #127 is in :+1:

PR #127, opened 2017-02-19
Validate Host header for HTTP/1.1 requests as per RFC

clue:
HTTP/1.1 requests MUST include a valid host header (host and optional port) as per RFC: https://tools.ietf.org/html/rfc7230#section-5.4

Builds on top of #124
Refs #125


clue:

PR #151, opened 2017-03-24
Simplify examples to ease getting started

clue:


andig:
I may be wrong but the examples are trying to listen on port 0?

clue:
@andig Yes, this is intended:

> In order to use a random port assignment, you can use the port 0:

https://github.com/reactphp/socket#server

andig:
@clue another thing learned, thank you!

PR #167, opened 2017-04-14
Use https-scheme for request URIs if secure TLS is used

clue:
`$request->getUri()->getScheme()` will now return `https` if secure TLS is used and `http` otherwise.

Builds on top of #146 and #158


clue:
Updated to work around PHP < 7 not giving access to stream crypto parameters :shipit:

Issue #192, opened 2017-05-26
Should the server params always be defined?


gsdevme:
http://www.faqs.org/rfcs/rfc3875.html

```
4.1.  Request Meta-Variables

   Meta-variables contain data about the request passed from the server
   to the script, and are accessed by the script in a system-defined
   manner.  Meta-variables are identified by case-insensitive names;
   there cannot be two different variables whose names differ in case
   only.  Here they are shown using a canonical representation of
   capitals plus underscore ("_").  A particular system can define a
   different representation.

      meta-variable-name = "AUTH_TYPE" | "CONTENT_LENGTH" |
                           "CONTENT_TYPE" | "GATEWAY_INTERFACE" |
                           "PATH_INFO" | "PATH_TRANSLATED" |
                           "QUERY_STRING" | "REMOTE_ADDR" |
                           "REMOTE_HOST" | "REMOTE_IDENT" |
                           "REMOTE_USER" | "REQUEST_METHOD" |
                           "SCRIPT_NAME" | "SERVER_NAME" |
                           "SERVER_PORT" | "SERVER_PROTOCOL" |
                           "SERVER_SOFTWARE" | scheme |
                           protocol-var-name | extension-var-name
      protocol-var-name  = ( protocol | scheme ) "_" var-name
      scheme             = alpha *( alpha | digit | "+" | "-" | "." )
      var-name           = token
      extension-var-name = token
```

Figs states the following about `getServerParams` 

```
     * Retrieves data related to the incoming request environment,
     * typically derived from PHP's $_SERVER superglobal. The data IS NOT
     * REQUIRED to originate from $_SERVER.
```

Im wondering if in the above they should at least exist or is that not the job of this package? 

I've made a PR to the package as they can guard for it also.. but Im wondering whats best also 
https://github.com/zendframework/zend-expressive/pull/489

clue:
Thanks for bringing this up! :+1: The first batch of server params has been implemented via #174 and I'm sure there's room for improvement if certain params are missing. May I ask you to file a PR so we can look into this? :shipit:

gsdevme:
Sure, are you happy with the idea of defining all in the meta variables in RFC3875?

clue:
Not sure what's reasonable here without diving deeper into this tbh. There's no need to implement this according to the CGI specs, but there are certainly a group of parameters not otherwise available or so commonly used that it makes sense to also add them here. I'd suggest going for these first :+1:

clue:
Closed via #195

PR #195, opened 2017-05-28
(WIP) feat(server-params): further define common server params

gsdevme:
Marking this is a WIP for now, need to further look at the best way to get `SCRIPT_FILENAME`.. somewhat might not be possible.. unless its passed as an option or defined elsewhere... if nothing else I think it should be set to an empty string though due to https://github.com/reactphp/http/issues/192

```
'SCRIPT_FILENAME'
The absolute pathname of the currently executing script.
```

- [ ] HTTP_HOST
- [ ] SCRIPT_FILENAME
- [ ] HTTP_USER_AGENT
- [ ] Should HTTPS be set to an empty string on non-https? (think so)
- [ ] Convert to long array Syntax to support an unsupported PHP version from 8 years ago

https://secure.php.net/manual/en/reserved.variables.server.php

### DIFF
--- a/src/RequestHeaderParser.php
+++ b/src/RequestHeaderParser.php
@@ -91,7 +91,12 @@ class RequestHeaderParser extends EventEmitter
         // previous properties and restoring original request-target
         $serverParams = array(
             'REQUEST_TIME' => time(),
-            'REQUEST_TIME_FLOAT' => microtime(true)
+            'REQUEST_TIME_FLOAT' => microtime(true),
+            'REQUEST_METHOD' => $request->getMethod(),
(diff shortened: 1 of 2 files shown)


clue:
Thank you for digging into this and filing this PR! :+1:

I still think that getting this feature in makes perfect sense and I'm looking forward to a future update here!

Given that this is currently a WIP and hasn't seen any updates in a while and I don't see it's likely this will get traction any time soon, I will have to close this for now. Please don't see this as discouraging, we would still love to get this feature in and very much appreciate your effort! You can simply complete this feature and update this PR and we're happy to reopen this :+1:

PR #205, opened 2017-07-04
Refactor to use HTTP response reason phrases from response object

clue:
This means we no longer have to maintain a list of HTTP reason phrases
ourselves, but instead rely on those already present in our HTTP message
abstraction.


src/Server.php
```diff
@@ -275,19 +275,22 @@ function ($error) use ($that, $conn, $request) {
...
         );
 
+        // append reason phrase to response body if known
+        $reason = $response->getReasonPhrase();
+        if ($reason !== '') {
+            $body = $response->getBody();
+            $body->seek(0, SEEK_END);
+            $body->write(': ' . $reason);
+        }
+
```
jsor: Any particular reason you don't construct `$message` as before?

```php
$message = 'Error ' . $code;

$reason = $response->getReasonPhrase();
if ($reason !== '') {
    $message .= ': ' . $reason
}
```

clue: The `$message` is used only as a constructor parameter for the response body and the default reason phrase is only available after calling the constructor, hence this assignment.

src/Server.php
```diff
@@ -275,19 +275,22 @@ function ($error) use ($that, $conn, $request) {
...
             ),
-            $message
+            'Error ' . $code
         );
 
+        // append reason phrase to response body if known
+        $reason = $response->getReasonPhrase();
+        if ($reason !== '') {
+            $body = $response->getBody();
+            $body->seek(0, SEEK_END);
```
kelunik: Why does it need that `seek`?

clue: This makes sure it seeks to the end so that we *append* the reason phrase to the existing body.

src/Server.php
jsor: Oh yes, missed that 😊

PR #208, opened 2017-07-19
Require Host on 1.1 requests

aaronbonneau:
[RFC 7230](https://tools.ietf.org/html/rfc7230#section-5.4) requires "A client MUST send a Host header field in all HTTP/1.1 request messages."

### DIFF
--- a/src/RequestHeaderParser.php
+++ b/src/RequestHeaderParser.php
@@ -163,6 +163,10 @@ class RequestHeaderParser extends EventEmitter
             }
         }
 
+        if ($request->getProtocolVersion() === '1.1' && !$request->hasHeader('Host')) {
+            throw new \InvalidArgumentException('A client must send a host header field in all HTTP/1.1 request messages');
+        }
+
         // Optional Host header value MUST be valid (host and optional port)
         if ($request->hasHeader('Host')) {
             $parts = parse_url('http://' . $request->getHeaderLine('Host'));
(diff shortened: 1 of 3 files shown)


src/RequestHeaderParser.php
```diff
@@ -163,6 +163,10 @@ private function parseRequest($headers)
             }
         }
 
+        if ($request->getProtocolVersion() === '1.1' && !$request->hasHeader('Host')) {
+            throw new \InvalidArgumentException('A client must send a host header field in all HTTP/1.1 request messages');
```
kelunik: Why `\InvalidArgumentException`?

aaronbonneau: The parameter to `parseRequest` (`$headers`) is not a valid argument.

Also, all exceptions in this method are `InvalidArgumentExceptions` so I was just keeping the format.

clue:
Thanks for filing this PR! :+1: I see where you're coming from and would love to hear some more thoughts on this!

I think the (more) relevant quote from the RFC would be this:

>    A server MUST respond with a 400 (Bad Request) status code to any
   HTTP/1.1 request message that lacks a Host header field and to any
   request message that contains more than one Host header field or a
   Host header field with an invalid field-value.

A similar patch was filed with #127 which was later reverted partly by #173 and #201 (all by me in fact). This was done both for consistency between HTTP/1.1 and HTTP/1.0 requests and because I've been running this project in production for a few months now and (despite violating RFC 7230) have repeatedly seen HTTP/1.1 requests without a `Host` header in production (https://github.com/leproxy/leproxy/pull/16).

The question actually boils down to: Is a `Host` header really required here and/or do we miss anything if we disregard the specs here?

clue:
Ping @aaronbonneau, what's the status here? :+1:

clue:
I'm closing this for now as it hasn't received any input in a while and I believe this has been answered. Please come back with more details if this problem persists and we can reopen this :+1:

PR #209, opened 2017-07-19
Fixed URI parsing #207

### LINKED ISSUE #207: Error when header parsing
When request target is origin-form and contain right absolute path, like this:
`GET /somepath?param=http://example.com HTTP/1.1`

`RequestHeaderParser` throw `InvalidArgumentException('Invalid absolute-form request-target')`

This error can be corrected by changing the test condition, for example
```
((!isset($parts['scheme'], $parts['host']) || $parts['scheme'] !== 'http' || isset($parts['fragment'])) && !isset($parts['path']))
```

aaronbonneau:
Closes #207

### DIFF
--- a/src/RequestHeaderParser.php
+++ b/src/RequestHeaderParser.php
@@ -154,11 +154,11 @@ class RequestHeaderParser extends EventEmitter
         }
 
         // ensure absolute-form request-target contains a valid URI
-        if (strpos($request->getRequestTarget(), '://') !== false) {
+        if (strpos($request->getRequestTarget(), '://') !== false && substr($request->getRequestTarget(), 0, 1) !== '/') {
             $parts = parse_url($request->getRequestTarget());
 
             // make sure value contains valid host component (IP or hostname), but no fragment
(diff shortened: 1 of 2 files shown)


src/RequestHeaderParser.php
```diff
@@ -158,7 +158,7 @@ private function parseRequest($headers)
             $parts = parse_url($request->getRequestTarget());
 
             // make sure value contains valid host component (IP or hostname), but no fragment
-            if (!isset($parts['scheme'], $parts['host']) || $parts['scheme'] !== 'http' || isset($parts['fragment'])) {
+            if ((!isset($parts['scheme'], $parts['host']) || $parts['scheme'] !== 'http' || isset($parts['fragment'])) && isset($parts['scheme'])) {
```
clue: This is now really hard to read and I'm not sure it does what it's supposed to do anymore. It looks like a `request-target` that starts with `://` will now pass this validation?

As an alternative, may I suggest reverting this and updating the above check instead? Your issue can only really happen for `origin-target` requests, which means we can actually skip this whole block if the `request-target` starts with a `/`.

Does this make sense to you?

aaronbonneau: Yes, this is very difficult to read; I'll update it.

Issue #231, opened 2017-10-03
Cannot use HttpServer with UDS


andig:
PHP-PM uses UDS for handling master requests on slave. This fails in the `RequestHeaderParser`:

	parser->error Symfony\Component\Debug\Exception\ContextErrorException: Notice: Undefined index: port in /Users/andig/Documents/htdocs/vz/vendor/react/http/src/RequestHeaderParser.php:107
	Stack trace:
	#0 /Users/andig/Documents/htdocs/vz/vendor/react/http/src/RequestHeaderParser.php(59): React\Http\RequestHeaderParser->parseRequest('GET /data.json ...')
	#1 /Users/andig/Documents/htdocs/vz/vendor/react/http/src/RequestHeaderParser.php(49): React\Http\RequestHeaderParser->parseAndEmitRequest(132)
	#2 [internal function]: React\Http\RequestHeaderParser->feed('GET /data.json ...')

ping @WyriHaximus can this constraint be removed to allow any type of underlying socket?

andig:
Note: for my testing purpose solved by:

        if ($this->remoteSocketUri !== null) {
            $remoteAddress = parse_url($this->remoteSocketUri);

            if (isset($remoteAddress['port'])) {
                $serverParams['REMOTE_ADDR'] = $remoteAddress['host'];
                $serverParams['REMOTE_PORT'] = $remoteAddress['port'];
            }
        }

Host is set to `________` in my tests, so testing for host does not work.

clue:
Thanks for filing this! This is an interesting issue, given that the underlying component does not currently support Unix domain sockets (UDS) (https://github.com/reactphp/socket/issues/25) :-)

I agree that despite this, this is still a minor bug that parsing the request should probably not fail in this case. Do you feel like filing a PR to fix (and test) this? :shipit: 

Considering long-term goals, does it make sense to ignore `REMOTE_ADDR` and `REMOTE_PORT` in this case and/or how do other servers handle this?

andig:
@clue I've added a test. The problem at least on OSX seems to be related to a PHP bug with UDS that I've seen before (https://bugs.php.net/bug.php?id=74556).

clue:
@andig Thanks for the update, I've just commented on this and linked to a related ticket (https://github.com/reactphp/http/pull/234#discussion_r143327946) :+1: 

[...]

PR #233, opened 2017-10-04
Return internal server error for unexpected exceptions

andig:
Debugging improvement for https://github.com/reactphp/http/issues/231

Alle exceptions are treated as HTTP 500 unless they explicitly provide an error code.

### DIFF
--- a/src/ChunkedDecoder.php
+++ b/src/ChunkedDecoder.php
@@ -95,7 +95,7 @@ class ChunkedDecoder extends EventEmitter implements ReadableStreamInterface
                 if ($positionCrlf === false) {
                     // Header shouldn't be bigger than 1024 bytes
                     if (isset($this->buffer[static::MAX_CHUNK_HEADER_SIZE])) {
-                        $this->handleError(new \Exception('Chunk header size inclusive extension bigger than' . static::MAX_CHUNK_HEADER_SIZE. ' bytes'));
+                        $this->handleError(new \Exception('Chunk header size inclusive extension bigger than' . static::MAX_CHUNK_HEADER_SIZE. ' bytes', 400));
                     }
                     return;
                 }
@@ -117,7 +117,7 @@ class ChunkedDecoder extends EventEmitter implements ReadableStreamInterface
 
                 $this->chunkSize = hexdec($hexValue);
                 if (dechex($this->chunkSize) !== $hexValue) {
-                    $this->handleError(new \Exception($hexValue . ' is not a valid hexadecimal number'));
+                    $this->handleError(new \Exception($hexValue . ' is not a valid hexadecimal number', 400));
                     return;
                 }
 
@@ -152,7 +152,7 @@ class ChunkedDecoder extends EventEmitter implements ReadableStreamInterface
 
             if ($positionCrlf !== 0 && $this->chunkSize === $this->transferredSize && strlen($this->buffer) > 2) {
                 // the first 2 characters are not CLRF, send error event
-                $this->handleError(new \Exception('Chunk does not end with a CLRF'));
+                $this->handleError(new \Exception('Chunk does not end with a CLRF', 400));
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


andig:
@clue if this behaviour change is what you'd like to see I'll go about correcting the test expectations.

src/Server.php
```diff
@@ -186,7 +186,7 @@ public function handleConnection(ConnectionInterface $conn)
 
             $that->writeError(
                 $conn,
-                $e->getCode() !== 0 ? $e->getCode() : 400
+                $e->getCode() !== 0 ? $e->getCode() : 500
```
kelunik: I don't think any exception code should just be forwarded to the client here as a status code. Most exception codes aren't HTTP status codes.

andig: > Most exception codes aren't HTTP status codes.

I've not been able to figure out if/what exception codes PHP uses. But inside react/http all codes are specifically generated with the purpose of being used as HTTP return codes.

andig:
The test failure

	1) React\Tests\Http\ServerTest::testRequestWithMalformedHostWillEmitErrorAndSendErrorResponse
	Failed asserting that 'HTTP/1.1 500 Internal Server Error

is interesting. It does not actually happen inside the `Server` code as might be expected here: https://github.com/reactphp/http/blob/master/src/RequestHeaderParser.php#L178

but inside ringcentral: https://github.com/reactphp/http/blob/master/src/RequestHeaderParser.php#L95

Now the question is if this should in the sense of this PR be a 400 or a 500?

src/Server.php
clue: This particular code only catches `Exception` instances from the `RequestHeaderParser` which is specifically designed to throw with HTTP status codes.

src/ChunkedDecoder.php
```diff
@@ -95,7 +95,7 @@ public function handleData($data)
                 if ($positionCrlf === false) {
                     // Header shouldn't be bigger than 1024 bytes
                     if (isset($this->buffer[static::MAX_CHUNK_HEADER_SIZE])) {
-                        $this->handleError(new \Exception('Chunk header size inclusive extension bigger than' . static::MAX_CHUNK_HEADER_SIZE. ' bytes'));
+                        $this->handleError(new \Exception('Chunk header size inclusive extension bigger than' . static::MAX_CHUNK_HEADER_SIZE. ' bytes', 400));
```
clue: While I agree that it may make sense for invalid request headers, I'm not sure about the changes in this class. This class only deals with decoding the request body which will be streamed to the request handler. As such, the status codes do not necessarily end up in the response code.

See also other occurrences in this class.

clue:
> Now the question is if this should in the sense of this PR be a 400 or a 500?

This is clearly triggered due to invalid client input and should thus reject with a `400` (Bad Request) status code instead of a `500` (Internal Server Error).

andig:
>> Now the question is if this should in the sense of this PR be a 400 or a 500?
> This is clearly triggered due to invalid client input and should thus reject with a 400 (Bad Request) status code instead of a 500 (Internal Server Error).

Fully agree for this test case. This would require catching the exception and re-throwing with status 400.

@clue Since we cannot really know what triggers a problem (e.g. the UDS issue)- do you think that this PR still makes sense or should I rather abandon it?

clue:
> @clue Since we cannot really know what triggers a problem (e.g. the UDS issue)- do you think that this PR still makes sense or should I rather abandon it?

Afaict we *do* know what can cause an `Exception` and there's no such thing as "unexpected exceptions". For instance, UDS (#231) should be fixed so that it no longer triggers an `E_WARNING` and/or `E_NOTICE`. This is apparently addressed in #234, so I'm not sure how much value this PR currently adds. If you happen to be able to trigger an "unexpected exception" in a test case, then I'm all for getting this in :+1:

andig:
Closing this for the time being.

PR #234, opened 2017-10-04
Prevent exception if underlying socket is a unix socket

andig:
Partial UDS support for https://github.com/reactphp/http/issues/231

### DIFF
--- a/src/RequestHeaderParser.php
+++ b/src/RequestHeaderParser.php
@@ -103,8 +103,12 @@ class RequestHeaderParser extends EventEmitter
 
         if ($this->remoteSocketUri !== null) {
             $remoteAddress = parse_url($this->remoteSocketUri);
-            $serverParams['REMOTE_ADDR'] = $remoteAddress['host'];
-            $serverParams['REMOTE_PORT'] = $remoteAddress['port'];
+
+            // port won't be set for unix domain sockets
+            if (isset($remoteAddress['port'])) {
+                $serverParams['REMOTE_ADDR'] = $remoteAddress['host'];
+                $serverParams['REMOTE_PORT'] = $remoteAddress['port'];
+            }
         }
 
         if ($this->localSocketUri !== null) {
--- a/tests/RequestHeaderParserTest.php
+++ b/tests/RequestHeaderParserTest.php
@@ -447,6 +447,18 @@ class RequestHeaderParserTest extends TestCase
         $this->assertEquals('this', $queryParams['test']);
     }
 
+    public function testUnixRemoteSocketWithoutPort()
+    {
+        // reactphp/socket Connect->parseAddress returns "tcp://NUL" for UDS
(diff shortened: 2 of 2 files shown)


src/RequestHeaderParser.php
```diff
@@ -103,8 +103,12 @@ private function parseRequest($headers)
...
         if ($this->remoteSocketUri !== null) {
             $remoteAddress = parse_url($this->remoteSocketUri);
-            $serverParams['REMOTE_ADDR'] = $remoteAddress['host'];
-            $serverParams['REMOTE_PORT'] = $remoteAddress['port'];
+
+            // port won't be set for unix domain sockets
+            if (isset($remoteAddress['port'])) {
+                $serverParams['REMOTE_ADDR'] = $remoteAddress['host'];
+                $serverParams['REMOTE_PORT'] = $remoteAddress['port'];
+            }
```
kelunik: But the `REMOTE_ADDR` should still be set, no? I guess the port could just be set to `0`.

andig: In case of OSX I get `________` as `REMOTE_ADDR` so I've removed both.

tests/RequestHeaderParserTest.php
```diff
@@ -447,6 +447,18 @@ public function testQueryParmetersWillBeSet()
         $this->assertEquals('this', $queryParams['test']);
     }
 
+    public function testUnixRemoteSocketWithoutPort()
+    {
+        // reactphp/socket Connect->parseAddress returns "tcp://NUL" for UDS
+        $parser = new RequestHeaderParser(null, 'tcp://' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '' . "\0" . '');
```
kelunik: Why these empty string concatenations?

andig: See https://bugs.php.net/bug.php?id=74556. Thats what PHP returns for stream address of UDS on OSX.

clue: @andig Thanks for the update!

Have you seen https://github.com/reactphp/socket/pull/100? IMO this should be handled by the Socket component, which is also where this address comes from (afaict). This way, we would likely expect a `null` address here. Does this make sense to you?

kelunik: Why do you want upvotes on an already fixed bug? But it's not what I asked. You write `"\0" . '' . "\0"` there instead of simply `"\0\0"`, why that?

andig: > You write "\0" . '' . "\0" there instead of simply "\0\0", why that?

That was copy/paste from `var_export`, will fix.

PR #235, opened 2017-10-09
Support buffering requests with unknown length (Transfer-Encoding: chunked) in RequestBodyBufferMiddleware

andig:
Implement chunked response handling, following discussion in https://github.com/reactphp/http/issues/232

### DIFF
--- a/README.md
+++ b/README.md
@@ -697,16 +697,11 @@ Any incoming request that has a request body that exceeds this limit will be
 rejected with a `413` (Request Entity Too Large) error message without calling
 the next middleware handlers.
 
-Any incoming request that does not have its size defined and uses the (rare)
-`Transfer-Encoding: chunked` will be rejected with a `411` (Length Required)
-error message without calling the next middleware handlers.
-Note that this only affects incoming requests, the much more common chunked
-transfer encoding for outgoing responses is not affected.
-It is recommended to define a `Content-Length` header instead.
-Note that this does not affect normal requests without a request body
-(such as a simple `GET` request).
+The `RequestBodyBufferMiddleware` will buffer requests with bodies of known size 
+(i.e. with `Content-Length` header specified) as well as requests with bodies of 
+unknown size (i.e. with `Transfer-Encoding: chunked` header).
 
-All other requests will be buffered in memory until the request body end has
+All requests will be buffered in memory until the request body end has
 been reached and then call the next middleware handler with the complete,
 buffered request.
 Similarly, this will immediately invoke the next middleware handler for requests
--- a/src/Middleware/RequestBodyBufferMiddleware.php
+++ b/src/Middleware/RequestBodyBufferMiddleware.php
@@ -53,7 +53,7 @@ final class RequestBodyBufferMiddleware
                 return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
             }
 
-            return $error;
+            throw $error;
         });
     }
 
--- a/src/Middleware/RequestBodyBufferMiddleware.php
+++ b/src/Middleware/RequestBodyBufferMiddleware.php
@@ -7,6 +7,7 @@ use React\Http\Response;
 use React\Promise\Stream;
 use React\Stream\ReadableStreamInterface;
 use RingCentral\Psr7\BufferStream;
+use OverflowException;
 
 final class RequestBodyBufferMiddleware
 {
@@ -29,27 +30,30 @@ final class RequestBodyBufferMiddleware
 
     public function __invoke(ServerRequestInterface $request, $stack)
     {
-        $size = $request->getBody()->getSize();
+        $body = $request->getBody();
 
-        if ($size === null) {
-            return new Response(411, array('Content-Type' => 'text/plain'), 'No Content-Length header given');
-        }
-
-        if ($size > $this->sizeLimit) {
+        // request body of known size exceeding limit
+        if ($body->getSize() > $this->sizeLimit) {
             return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
         }
 
-        $body = $request->getBody();
         if (!$body instanceof ReadableStreamInterface) {
             return $stack($request);
         }
 
-        return Stream\buffer($body)->then(function ($buffer) use ($request, $stack) {
+        return Stream\buffer($body, $this->sizeLimit)->then(function ($buffer) use ($request, $stack) {
             $stream = new BufferStream(strlen($buffer));
             $stream->write($buffer);
             $request = $request->withBody($stream);
 
             return $stack($request);
+        }, function($error) {
+            // request body of unknown size exceeding limit during buffering
+            if ($error instanceof OverflowException) {
+                return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
+            }
+
+            return $error;
         });
     }
 
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "react/stream": "^1.0 || ^0.7.1",
         "react/promise": "^2.3 || ^1.2.1",
         "evenement/evenement": "^3.0 || ^2.0 || ^1.0",
-        "react/promise-stream": "^0.1.1"
+        "react/promise-stream": "^1.0 || ^0.1.2"
     },
     "autoload": {
         "psr-4": {
--- a/tests/Middleware/RequestBodyBufferMiddlewareTest.php
+++ b/tests/Middleware/RequestBodyBufferMiddlewareTest.php
@@ -3,12 +3,14 @@
 namespace React\Tests\Http\Middleware;
 
 use Psr\Http\Message\ServerRequestInterface;
+use React\EventLoop\Factory;
 use React\Http\Io\HttpBodyStream;
 use React\Http\Io\ServerRequest;
 use React\Http\Middleware\RequestBodyBufferMiddleware;
 use React\Stream\ThroughStream;
 use React\Tests\Http\TestCase;
 use RingCentral\Psr7\BufferStream;
+use Clue\React\Block;
 
 final class RequestBodyBufferMiddlewareTest extends TestCase
 {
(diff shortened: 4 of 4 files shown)


andig:
@WyriHaximus I need your help with rebasing this for https://github.com/reactphp/promise-stream/pull/3. In the tests there's currently this for the `RequestBodyBufferMiddlewareTest`:

    public function testExcessiveSizeReturnsError413()
    {
        $stream = new BufferStream(2);
        $stream->write('aa');

        $serverRequest = new ServerRequest(
            'GET',
            'https://example.com/',
            array(),
            $stream
        );

        $buffer = new RequestBodyBufferMiddleware(1);
        $response = $buffer(
            $serverRequest,
            function () {}
        );

        $this->assertSame(413, $response->getStatusCode());
    }

Now what goes in is NOT a `ReadableStreamInterface` and therefore returned early before any buffering in https://github.com/reactphp/http/blob/master/src/Middleware/RequestBodyBufferMiddleware.php#L43. 

Is that really desired, i.e. should it be possible to construct `ServerRequest` with a body that doesn't implement Psr7 semantics (and therefore actually is a `HttpBodyStream`? Hidden bug or feature that I fail to grasp?

clue:
@andig Have you seen the documentation for the [`RequestBodyBufferMiddleware`](https://github.com/reactphp/http#requestbodybuffermiddleware)? I think that this middleware should support all valid requests. In a normal stack this will be the incoming streaming request, but another prior middleware handler may also expose a buffered request to this middleware handler.

The following (completely contrived) example should allow request up to 8 MiB:
```
$middleware = new MiddlewareRunner([
    new RequestBodyBufferMiddleware(16 * 1024 * 1024), // 16 MiB
    new RequestBodyBufferMiddleware(8 * 1024 * 1024), // 8 MiB
});
```

src/Middleware/RequestBodyBufferMiddleware.php
```diff
@@ -29,27 +30,23 @@ public function __construct($sizeLimit = null)
...
     {
-        $size = $request->getBody()->getSize();
-
-        if ($size === null) {
-            return new Response(411, array('Content-Type' => 'text/plain'), 'No Content-Length header given');
-        }
-
-        if ($size > $this->sizeLimit) {
-            return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
-        }
```
clue: We should probably keep this check: it makes sense to fail early in case the client is trying to upload a gigabyte file over a slow uplink :-)

andig: Fixed. Please forgive me for force-pushing. I didn't realize you were already reviewing.

andig:
Small side note: I've also changed mentions of `g7` and `Psr7Implemention` to `Psr7` for sake of coding consistency.

andig:
@clue 

> Have you seen the documentation for the `RequestBodyBufferMiddleware`? I think that this middleware should support all valid requests.

Reading this:

> This can be useful if full PSR-7 compatibility is needed for the request handler and the default streaming request body handling is not needed.

I understand the behaviour. It's still totally confusing though that- if you use the `RequestBodyBufferMiddleware` and send in a react stream with valid body it can be of any size and will *not* be buffered or rejected.

Shouldn't at least the `RequestBodyBufferMiddleware` reject non-Psr7 bodies as they are used in the tests?

src/Middleware/RequestBodyBufferMiddleware.php
clue: No worries, happy to see you're faster with coding than I am with reviewing :shipit:

clue:
I'm not sure I follow tbh. Our middleware implementation works around the concept of PSR-7 requests coming in and PSR-7 responses coming out (plus wrapped in promises). This means that there's no such thing as a "non-Psr7 body".

What's special about this implementation is that bodies can be in a streaming state when a `ReadableStreamInterface` is attached. This allows us to transport bodies of arbitrary sizes without having to store huge bodies in memory.

Does this help?

andig:
> I'm not sure I follow tbh.

Never mind. I've just noticed that I had to rewrite https://github.com/reactphp/http/pull/235/files#diff-5ff94d7c5e50a34ed39b11c33717c0b9R68 to use a different stream type. I'm probably making this too complicated ;)

WyriHaximus:
@andig Seems you're getting there, keep up the good work :+1:

src/HttpBodyStream.php
```diff
@@ -25,7 +25,7 @@ class HttpBodyStream extends EventEmitter implements StreamInterface, ReadableSt
      * @param ReadableStreamInterface $input - Stream data from $stream as a body of a PSR-7 object4
      * @param int|null $size - size of the data body
      */
-    public function __construct(ReadableStreamInterface $input, $size)
+    public function __construct(ReadableStreamInterface $input, $size = null)
```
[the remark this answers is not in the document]
andig: Imho this change makes sense. It's aligned with the comment now, the constructor handles null and it prevents explicitly passing null for bodies of unknown length?

src/RequestHeaderParser.php
```diff
@@ -92,7 +92,7 @@ private function parseRequest($headers)
         }
 
         // parse request headers into obj implementing RequestInterface
-        $request = g7\parse_request($headers);
+        $request = Psr7\parse_request($headers);
```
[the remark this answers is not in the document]
andig: I wanted to have Psr7 consistently named in the codebase (not Psr7, g7, and PsrImplementation). Could move to separate PR?

src/Response.php
```diff
@@ -21,7 +21,7 @@ public function __construct(
         $version = '1.1',
         $reason = null
     ) {
-        if ($body instanceof ReadableStreamInterface) {
+        if ($body instanceof ReadableStreamInterface &! $body instanceof HttpBodyStream) {
```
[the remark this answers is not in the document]
andig: Artifact removed.

src/Server.php
```diff
@@ -3,13 +3,12 @@
 namespace React\Http;
```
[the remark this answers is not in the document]
andig: Will open separate PR for Psr7 but imho at least this change https://github.com/reactphp/http/pull/235/files#diff-c611f5024b5b13f69f779505d1c6735eR170 makes sense.

Do you mind the reordering of the use clauses by library?

src/HttpBodyStream.php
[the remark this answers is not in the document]
clue: I feel that this adds noise and adds no value and as part of this PR, maybe suggest this in a separate PR? :+1: 

Despite this, I don't see what value this would provide, as this is an internal class only and this change basically means you're allowed th instantiate the class without considering what its "size" would be. If you feel this is relevant, I encourage you to file a PR so we can discuss this further :+1:

WyriHaximus:
@andig :tada: https://github.com/reactphp/promise-stream/releases/tag/v0.1.2 :tada:

WyriHaximus:
@andig updated this PR's title to something that reflects what the current code is doing

andig:
I've added one more commit to test for 413 in case on known but excessive content length. Still need to have another look at the README.

WyriHaximus:
@andig could you update to `^1.0 || ^0.1.2` since `1.0` has been release earlier today? :tada:

README.md
```diff
@@ -697,16 +697,17 @@ Any incoming request that has a request body that exceeds this limit will be
...
-transfer encoding for outgoing responses is not affected.
-It is recommended to define a `Content-Length` header instead.
-Note that this does not affect normal requests without a request body
-(such as a simple `GET` request).
+The `RequestBodyBufferMiddleware` will buffer requests with bodies of known size 
+(i.e. with `Content-Length` header specified) as well as requests with bodies of 
+unknown size (i.e. with `Transfer-Encoding: chunked` header). To simplify request 
+handling for downstream middlewares or applications buffered `chunked` requests will 
+have the `Transfer-Encoding` header removed and a `Content-Length` header with the 
+actual request body size received added instead.
```
clue: README updates are much appreciated :+1: However, I feel this is a bit misleading here, as this makes it sounds like this is what this middleware handler does, whereas this is actually handled by the `Server` itself. Also, the word "middlewares" doesn't exist (think "softwares"), I usually use "middleware handlers" instead :-) Can you rephrase this? :+1:

README.md
```diff
@@ -697,16 +697,17 @@ Any incoming request that has a request body that exceeds this limit will be
...
+
+Note that this only affects incoming requests, the chunked transfer encoding for 
+outgoing responses is not affected.
```
clue: This sentence made much more sense for the old behavior where these requests were rejected. Does it make sense to you to remove this sentence?

README.md
andig: I think it has actually become a duplicate since the response section already describes the behaviour. Removed.

README.md
andig: Yep, removed.

src/Middleware/RequestBodyBufferMiddleware.php
```diff
@@ -29,27 +30,30 @@ public function __construct($sizeLimit = null)
...
+        $body = $request->getBody();
 
-        if ($size > $this->sizeLimit) {
+        // request body of known size exceeding limit
+        if ($body->getSize() > $this->sizeLimit) {
             return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
         }
 
-        $body = $request->getBody();
         if (!$body instanceof ReadableStreamInterface) {
```
andig: Shouldn't this check be moved up as this type of stream won't be buffered anyway? Why complain about size in this case (shouldn't happen when using the server at all, might as well just throw)?

clue: IMO it makes sense to keep this like this in (the rare) case this middleware is used behind another buffering middleware handler, see also https://github.com/reactphp/http/pull/235#issuecomment-336528451 :+1:

src/Middleware/RequestBodyBufferMiddleware.php
```diff
@@ -29,27 +30,28 @@ public function __construct($sizeLimit = null)
...
+        return Stream\buffer($body, $this->sizeLimit)->then(function ($buffer) use ($request, $stack) {
             $stream = new BufferStream(strlen($buffer));
             $stream->write($buffer);
             $request = $request->withBody($stream);
 
             return $stack($request);
+        }, function($error) {
+            if ($error instanceof OverflowException) {
+                return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
+            }
```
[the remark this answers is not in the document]
clue: I think this is a relevant point, let's postpone this discussion for now and address this in a follow-up ticket :shipit: :+1: 

See for example https://stackoverflow.com/questions/14250991/is-it-acceptable-for-a-server-to-send-a-http-response-before-the-entire-request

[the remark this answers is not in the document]
andig: Understood. Looking forward to your insights once this is in.

src/Middleware/RequestBodyBufferMiddleware.php
andig: @clue What I mean is that- if not a `ReadableStreamInterface`, which means we won't buffer it, we're still returning 413 although the middleware is _not_ responsible for handling this request as it is right now.

My suggestion would be to move the `ReadableStreamInterface` to the Beginning of the method?

andig: ping @clue could you check my question above? 
What's missing to make make progress with this PR (and 0.8 in general)?

clue: Well, technically buffering and limiting sizes are in fact independent responsibilities. For practical reasons however, I think that it makes sense to bring them close together because buffering without applying a limit is actually harmful (potential for trivial DOS otherwise). If we accept the fact that checking size limits here, then I think it makes sense to also support buffered requests here. This means that I think it makes sense to also reject buffered requests that exceed the given size limits.

Accordingly, I think this code LGTM :+1: What do you think about this?

src/Middleware/RequestBodyBufferMiddleware.php
```diff
@@ -29,27 +30,30 @@ public function __construct($sizeLimit = null)
...
             $request = $request->withBody($stream);
 
             return $stack($request);
+        }, function($error) {
+            // request body of unknown size exceeding limit during buffering
+            if ($error instanceof OverflowException) {
+                return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
+            }
+
+            return $error;
```
clue: This probably shouldn't `return` an `Exception` here :-) Can you fix this and add a test?

andig: @clue good catch! Undecided what to return though- right now it's an HTTP 400 with the Exception message as text. Do you have a better suggestion?

src/Middleware/RequestBodyBufferMiddleware.php
andig: > What do you think about this?

@clue sratch head. I'm not getting my point across. We're only buffering and limiting PSR streams. We are however *still* limiting react-only streams. 

This is imho wrong- as the buffering middleware is not responsible for these requests. I'm proposing-  to be explicit- to change order of things like this:

    public function __invoke(ServerRequestInterface $request, $stack)
    {
        $body = $request->getBody();

        if (!$body instanceof ReadableStreamInterface) {
            return $stack($request);
        }

        // request body of known size exceeding limit
        if ($body->getSize() > $this->sizeLimit) {
            return new Response(413, array('Content-Type' => 'text/plain'), 'Request body exceeds allowed limit');
        }

clue: Yes, it looks like we're talking past one another :man_shrugging: :-)

The `getBody()` method is guaranteed to *always* return an instance of `Psr\Message\StreamInterface`. This very instance *may* also implement the `React\Stream\ReadableStreamInterface` ("streaming state"), but this is not guaranteed (already in "buffered state").

When this middleware handler is used as documented, then the body is always in a "streaming state". If one were to use this middleware handler twice in the same stack (conceived example), then the first will see a body in "streaming state" while the second will see it in "buffered state".

If this conceived(!) second middleware handler uses a smaller size limit (which I would expect to be more of a programming error than intentional), then IMO the second middleware handler should still reject the message in my opinion.

If I understand your above suggestion correctly, then it looks like this would actually allow the body to pass through without applying the second limit at all.

I think it makes sense to add a test for this behavior to ensure that this is properly covered. Can you add a test for this or would you rather want me to add this to this PR? :+1:

src/Middleware/RequestBodyBufferMiddleware.php
andig: Done. Do you think we need to guard for error actually being an exception and not something else?

WyriHaximus:
What is the status here @clue and @andig?

andig:
Needs review of the last 2 commits

PR #237, opened 2017-10-11
Add MiddlewareRunner test ensuring multiple runs always invoke all middleware

jsor:
Ref https://github.com/reactphp/http/pull/236#issuecomment-335716226

### DIFF
--- a/tests/MiddlewareRunnerTest.php
+++ b/tests/MiddlewareRunnerTest.php
@@ -3,6 +3,7 @@
 namespace React\Tests\Http;
 
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
 use React\EventLoop\Factory;
 use React\Http\MiddlewareRunner;
 use React\Http\ServerRequest;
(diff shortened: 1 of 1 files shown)


tests/TestCase.php
```diff
@@ -59,6 +59,27 @@ protected function expectCallableConsecutive($numberOfCalls, array $with)
...
+        for ($i = 0, $l = count($with); $i < $l; $i++) {
+            $mock
+                ->expects($this->at($i))
+                ->method('__invoke')
+                ->with($this->equalTo($with[$i]))
+                ->willReturnCallback($returnCallback[$i]);
+        }
+
+        return $mock;
+    }
```
clue: How would we test this method? :yum: 

On a more serious note, this looks really complicated for such a simple test. As an alternative, maybe keep a counter somewhere and explicitly call the handler twice?

jsor: This is a copy of the `expectCallableConsecutive()` from above with additional `$mock->willReturnCallback()` ;)

> As an alternative, maybe keep a counter somewhere and explicitly call the handler twice?

Not sure what you mean here? Could you elaborate?

jsor: @clue I've updated the test and remove the test helper.

PR #274, opened 2017-12-08
Respect ini settings for file uploads

### LINKED ISSUE #257: RequestBodyParserMiddleware should respect ini settings

clue:
This PR ensures that the `RequestBodyParserMiddleware` respects all relevant ini settings for file uploads. In other words, this middleware will now respect the following ini settings:

```
file_uploads		On
max_file_uploads	20
upload_max_filesize	2M
```

Additionally, two parameters have been added to explicitly configure these settings like this:

```php
// allow up to 20 files with 2 MiB each
new RequestBodyParserMiddleware(2 * 1024 * 1024, 20);
```

I realize that this PR is not exactly trivial. If you want to review this, I would suggest checking out the individual commits.

Builds on top of #268
Resolves / closes #257

### DIFF
--- a/src/Io/MultipartParser.php
+++ b/src/Io/MultipartParser.php
@@ -56,7 +56,10 @@ final class MultipartParser
 
     private $postCount = 0;
 
-    public function __construct()
+    /**
+     * @param int|null $uploadMaxFilesize
+     */
+    public function __construct($uploadMaxFilesize = null)
     {
         $var = ini_get('max_input_vars');
         if ($var !== false) {
@@ -67,7 +70,7 @@ final class MultipartParser
             $this->maxInputNestingLevel = (int)$var;
         }
 
-        $this->uploadMaxFilesize = $this->iniUploadMaxFilesize();
+        $this->uploadMaxFilesize = $uploadMaxFilesize === null ? $this->iniUploadMaxFilesize() : (int)$uploadMaxFilesize;
     }
 
     public function parse(ServerRequestInterface $request)
--- a/src/Io/MultipartParser.php
+++ b/src/Io/MultipartParser.php
@@ -47,6 +47,13 @@ final class MultipartParser
      */
     private $maxInputNestingLevel = 64;
 
+    /**
+     * ini setting "upload_max_filesize"
+     *
+     * @var int
+     */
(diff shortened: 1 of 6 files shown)


src/Io/MultipartParser.php
```diff
@@ -67,18 +81,29 @@ private function __construct(ServerRequestInterface $request)
         if ($var !== false) {
             $this->maxInputNestingLevel = (int)$var;
         }
+
+        $this->uploadMaxFilesize = $uploadMaxFilesize === null ? $this->iniUploadMaxFilesize() : (int)$uploadMaxFilesize;
```
jsor: Does is make sense to also allow a string similar to the ini setting, like `"8M"`? The conversion logic is already [there](https://github.com/reactphp/http/pull/274/files#diff-e583461d932d2fbc3b560b92c2eeb8adR333).

src/Middleware/RequestBodyParserMiddleware.php
```diff
@@ -7,6 +7,16 @@
 
 final class RequestBodyParserMiddleware
 {
+    private $multipart;
+
+    /**
+     * @param int|null $uploadMaxFilesize
+     */
```
jsor: `@param int|null $maxFileUploads`missing?

src/Io/MultipartParser.php
clue: Good point! I'm not opposed, but would rather look into this in a follow-up PR because this also affects the `RequestBodyBufferMiddleware` :+1:

src/Middleware/RequestBodyParserMiddleware.php
clue: Good catch, updated :shipit:

src/Io/MultipartParser.php
jsor: Opened #275 for that 👍

PR #277, opened 2017-12-10
Simplify using middleware request handlers

clue:
We now support middleware request handlers by default and no longer require an explicit call to `MiddlewareRunner`. Accordingly, the `MiddlewareRunner` is now internal only.

Builds on top of #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,19 +659,41 @@ passed explicitly.
 
 ### Middleware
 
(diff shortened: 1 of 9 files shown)


jsor:
Test failing after merging to master. Not sure why now and not in the PR build. Opened #279 with a fix.

PR #279, opened 2017-12-11
Add missing use statement for MiddlewareRunner 

jsor:
Add missing use statement for MiddlewareRunner after moving to Io subnamespace.

Follow-up for #277


andig:
Was just gonna report this: https://travis-ci.org/php-pm/php-pm/jobs/314626879 but you're faster :)

:shipit:

PR #285, opened 2017-12-12
Both Server and StreamingServer accept middleware arrays

clue:
Builds on top of #277 and #281


clue:
Minor update to avoid failing test on legacy PHP 5.3 :shipit:

PR #292, opened 2017-12-30
Improve performance by avoiding unneeded promise wrapping

clue:
This very simple PR avoids wrapping the return value from the request handler in another promise. The request handler will either return a promise or a value that can be consumed directly. When running example 99 with included benchmarking instructions, this simple change improves performance from ~2400 req/s to ~2500 req/s on my local machine.

Builds on top of #291

### DIFF
--- a/src/StreamingServer.php
+++ b/src/StreamingServer.php
@@ -14,8 +14,9 @@ use React\Http\Io\LengthLimitedStream;
 use React\Http\Io\MiddlewareRunner;
 use React\Http\Io\RequestHeaderParser;
 use React\Http\Io\ServerRequest;
+use React\Promise;
 use React\Promise\CancellablePromiseInterface;
-use React\Promise\Promise;
+use React\Promise\PromiseInterface;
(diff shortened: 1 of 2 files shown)


src/StreamingServer.php
```diff
@@ -229,22 +230,43 @@ public function handleRequest(ConnectionInterface $conn, ServerRequestInterface
...
+            $stream->close();
+        }
+
+        // happy path: response returned, handle and return immediately
+        if ($response instanceof ResponseInterface) {
+            return $this->handleResponse($conn, $request, $response);
+        }
+
+        // did not return a promise? this is an error, convert into one for rejection below.
+        if (!$response instanceof PromiseInterface) {
```
jsor: This `if` should probably omitted and promise conversion should be completely delegated to `resolve()`.

clue: I'm not opposed to your suggestion, but after trying this out locally, the performance improvement is much less significant. Do you feel this is more of a cosmetic issue or would this block these suggested changes?

PR #295, opened 2018-01-03
StreamingServer::handleConnection bug fix

PabloJoan:
also fix problematic Io/IniUtil::iniSizeToBytes function

### DIFF
--- a/src/Io/IniUtil.php
+++ b/src/Io/IniUtil.php
@@ -10,13 +10,13 @@ final class IniUtil
     /**
      * Convert a ini like size to a numeric size in bytes.
      *
-     * @param string $size
-     * @return int
+     * @param number $size
+     * @return number
      */
     public static function iniSizeToBytes($size)
     {
         if (is_numeric($size)) {
-            return (int)$size;
+            return $size;
         }
 
         $suffix = strtoupper(substr($size, -1));
@@ -31,18 +31,18 @@ final class IniUtil
         }
 
         if ($suffix === 'K') {
-            return (int)($strippedSize * 1024);
+            return $strippedSize * 1024;
         }
         if ($suffix === 'M') {
-            return (int)($strippedSize * 1024 * 1024);
+            return $strippedSize * 1024 * 1024;
         }
         if ($suffix === 'G') {
-            return (int)($strippedSize * 1024 * 1024 * 1024);
+            return $strippedSize * 1024 * 1024 * 1024;
         }
         if ($suffix === 'T') {
-            return (int)($strippedSize * 1024  * 1024 * 1024 * 1024);
+            return $strippedSize * 1024  * 1024 * 1024 * 1024;
         }
 
-        return (int)$size;
+        return $size;
     }
 }
--- a/src/Io/IniUtil.php
+++ b/src/Io/IniUtil.php
@@ -10,13 +10,13 @@ final class IniUtil
     /**
      * Convert a ini like size to a numeric size in bytes.
      *
-     * @param string $iniSetting
+     * @param string $size
      * @return int
      */
     public static function iniSizeToBytes($size)
     {
         if (is_numeric($size)) {
-            return $size;
+            return (int)$size;
         }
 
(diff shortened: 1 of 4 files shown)


src/Io/IniUtil.php
```diff
@@ -10,13 +10,13 @@
     /**
      * Convert a ini like size to a numeric size in bytes.
      *
-     * @param string $iniSetting
+     * @param string $size
      * @return int
```
PabloJoan: the point of this change is to assure @return int remains true

src/Io/IniUtil.php
```diff
@@ -10,13 +10,13 @@
...
      */
     public static function iniSizeToBytes($size)
     {
         if (is_numeric($size)) {
-            return $size;
+            return (int)$size;
```
PabloJoan: the previous would return a numeric string, not an int

src/Io/IniUtil.php
```diff
@@ -31,18 +31,18 @@ public static function iniSizeToBytes($size)
         }
 
         if ($suffix === 'K') {
-            return $strippedSize * 1024;
+            return (int)($strippedSize * 1024);
```
PabloJoan: the previous could potentially return a float, not an int

src/Io/MultipartParser.php
```diff
@@ -101,7 +101,6 @@ public function parse(ServerRequestInterface $request)
         $this->parseBody('--' . $matches[1], (string)$request->getBody());
 
         $request = $this->request;
-        $this->request = null;
```
PabloJoan: ServerRequestInterface::$request does not accept null

andig: > ServerRequestInterface::$request does not accept null

This is about `MultipartParser->$request` which can be null and not related to parameter passing?

src/Io/IniUtil.php
clue: Good catch, but see also below for possible overflow :+1:

src/Io/IniUtil.php
clue: Note that an integer overflow will automatically cast to float in PHP. This means that `8G` on a 32 bit system will return the correct float value, while your change enforces an int and thus limits this value to` PHP_INT_MAX` again. I'm not sure how relevant this is here, but certainly something we should be aware of.

How about using a `number` type hint instead here? http://php.net/manual/de/language.pseudo-types.php#language.types.number

src/Io/MultipartParser.php
clue: Good catch, but with this change applied this instance will now keep a reference to the request and will thus avoid garbage collecting this instance until this method is called the next time.

How about using a `?ServerRequestInterface` type hint for this property instead? This also makes sense because the constructor does not assign an instance to this property.

src/StreamingServer.php
```diff
@@ -171,7 +171,7 @@ public function handleConnection(ConnectionInterface $conn)
         $parser = new RequestHeaderParser($uriLocal, $uriRemote);
 
         $listener = array($parser, 'feed');
-        $parser->on('headers', function (RequestInterface $request, $bodyBuffer) use ($conn, $listener, $that) {
+        $parser->on('headers', function (ServerRequestInterface $request, $bodyBuffer) use ($conn, $listener, $that) {
```
clue: Good catch, but should also remove the now unneeded import? :+1:

src/Io/IniUtil.php
clue: See below, maybe use `number` type hint here? http://php.net/manual/de/language.pseudo-types.php#language.types.number

src/StreamingServer.php
PabloJoan: removed

PabloJoan:
All changes have been made. Thanks for reviewing!

clue:
@PabloJoan Thank you for keeping up with this! I've cherry-picked your relevant changes (with full credits) to #301 because it also addresses similar issues. I'm closing this one in favor of the other one to finally get this merged :shipit:

PR #300, opened 2018-01-16
Last middleware can invoke next

jsor:
Allows the last middleware in MiddlewareRunner to invoke `$next()`.

Example

```php
$middleware = new MiddlewareRunner(array(
    function (RequestInterface $request, $next) {
        return $next($request);
    }
));
```
 See eg. https://travis-ci.org/jsor-labs/http/jobs/329394945 for a failing test.

### DIFF
--- a/src/Io/MiddlewareRunner.php
+++ b/src/Io/MiddlewareRunner.php
@@ -48,6 +48,10 @@ final class MiddlewareRunner
             return $that->call($request, $position + 1);
         };
 
+        if (!isset($this->middleware[$position])) {
+            return null;
+        }
+
         $handler = $this->middleware[$position];
         return $handler($request, $next);
     }
--- a/tests/Io/MiddlewareRunnerTest.php
+++ b/tests/Io/MiddlewareRunnerTest.php
@@ -454,4 +454,19 @@ final class MiddlewareRunnerTest extends TestCase
         $this->assertTrue($promise instanceof CancellablePromiseInterface);
         $promise->cancel();
     }
+
+    public function testLastMiddlewareCanInvokeNext()
+    {
+        $middleware = new MiddlewareRunner(array(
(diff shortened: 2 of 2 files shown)


src/Io/MiddlewareRunner.php
```diff
@@ -48,6 +48,10 @@ public function call(ServerRequestInterface $request, $position)
             return $that->call($request, $position + 1);
         };
 
+        if (!isset($this->middleware[$position])) {
+            return null;
```
clue: Not sure about returning `null` here as this would break the middleware specification (see README).

It's my understanding that this would be an error, so it should probably `throw` instead (php.net/manual/en/class.outofrangeexception.php)?

As an alternative, we *could* also use this to create a "default response", i.e. some kind of response factory and just return a "default response" here?

jsor: My understanding is, that it should behave like the middleware just didn't return anything because `$next()` invoked from the last middleware is just no-op since there is no *next*.

```php
$middleware = new MiddlewareRunner(array(
    function (RequestInterface $request, $next) {
        // no-op
    }
));
```
Since the MiddlewareRunner is internal and `null` will produce an [error](https://github.com/reactphp/http/blob/master/src/StreamingServer.php#L263), i think that this is an acceptable behaviour.

If we throw, we probably should explicitly document, that calling `$next()` from the last middleware is forbidden or just not pass `$next()` to the last middleware.

clue: > […] there is no next. […] just not pass $next() to the last middleware.

Now that you've said this, I think this actually makes most sense. The last request handler in the chain should not be allowed to call the "next".

clue: I think you're raising a very good point here. I've just filed #308 to implement this as suggested. Let me know what you think :+1:

PR #301, opened 2018-01-16
Some code improvements

seregazhuk:
This PR provides some small code improvements. It:
- removes unused variables
- simplifies statements
- removes unnecessary properties initialization

### DIFF
--- a/src/Io/PauseBufferStream.php
+++ b/src/Io/PauseBufferStream.php
@@ -29,7 +29,7 @@ class PauseBufferStream extends EventEmitter implements ReadableStreamInterface
     private $dataPaused = '';
     private $endPaused = false;
     private $closePaused = false;
-    private $errorPaused = null;
+    private $errorPaused;
     private $implicit = false;
 
     public function __construct(ReadableStreamInterface $input)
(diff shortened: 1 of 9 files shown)


src/Io/PauseBufferStream.php
```diff
@@ -29,7 +29,7 @@ class PauseBufferStream extends EventEmitter implements ReadableStreamInterface
     private $dataPaused = '';
     private $endPaused = false;
     private $closePaused = false;
-    private $errorPaused = null;
+    private $errorPaused;
```
clue: I understand that these assignments are essentially a NOOP. I've intentionally added these in the past to explicitly state that this property isn't initialized as part of the constructor (see also other occurrences). Would love to hear some more thoughts on this, but won't block this either way. :+1:

clue:
For the reference: This is also somewhat related to #295, does it make sense to cherry-pick (some of) the changes into a single PR?

src/Io/PauseBufferStream.php
seregazhuk: As for me, I think that the constructor itself says more explicitly about it, rather then assigned `null` default value. The idea behind these assignments wasn't clear for me, so I've considered them as useless. But if it is a common practice in the code, of course I will undo this change.

seregazhuk:
Sorry, but I'm not sure, that I know how to cherry pick commit from my request to someone's else 😞

clue:
I've just cherry picked @PabloJoan's relevant changes from #295 to this PR :shipit:

PR #310, opened 2018-03-22
Documentation improvement

seregazhuk:
This PR is the same as #297 
Removes word `function` from middleware description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -860,7 +860,7 @@ $server = new Server(array(
 
 Similarly, you can use the result of the `$next` middleware request handler
 function to modify the outgoing response.
-Note that as per the above documentation, the `$next` function may return a
+Note that as per the above documentation, the `$next` middleware request handler may return a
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -860,7 +860,7 @@ $server = new Server(array(
 
 Similarly, you can use the result of the `$next` middleware request handler
 function to modify the outgoing response.
-Note that as per the above documentation, the `$next` function may return a
+Note that as per the above documentation, the `$next` may return a
```
jsor: "...the `$next` may return a..." reads a bit odd imho.

seregazhuk: May be "the `$next` variable" or "the callable `$next`"?

jsor: In the other paragraph ↓, its called "`$next` middleware request handler", maybe that?

seregazhuk: Fixed.

PR #333, opened 2019-02-13
Add verify_peer SSL context

woytam:
For standard HTTPS server, it is necessary to disable peer verification, which is only used when client are authenticated by client certificates. HTTPS is used only server certificates.


clue:
I'll assume this is resolved and will close this for now, please feel free to report back otherwise :+1:

PR #335, opened 2019-02-20
Note about limiting concurency in Server

woytam:
For better understanding of not overwriting LimitConcurency in `Server`

### DIFF
--- a/README.md
+++ b/README.md
@@ -950,6 +950,10 @@ resumed).
 The following example shows how this middleware can be used to ensure no more
 than 10 handlers will be invoked at once:
 
+> Note, that the real concurency limit can be lower due to
+  [internal calculations](src/Server.php#L58) in [`Server`](#server)
+  based on `memory_limit` and `post_max_size` ini settings.
+
 ```php
 $server = new Server(array(
     new LimitConcurrentRequestsMiddleware(10),


andyvanee:
I just encountered the same issue, as I was seeing no speedups after 4 concurrent requests. I was using the php docker container which uses the default ini settings of `memory_limit = 128M` and `post_max_size = 8M` which results in only 4 concurrent requests.

https://github.com/php/php-src/blob/master/php.ini-development#L401

Since I'm not really handling any uploads, I added the following .ini file, which boosted it to 4096 concurrent requests.

```ini
[PHP]

post_max_size = 8K
memory_limit = 256M
```

The note in in the last paragraph of the [Server section](https://github.com/reactphp/http#server) of the README does cover this already, but I was confused because the automatically added middleware was setting the upper bound.

I am curious about how to calculate a sensible number for concurrent requests and what the failure mode is if I push it too far? I guess I may have to do some more testing 😃

clue:
@woytam Thank you for filing this PR, I agree that it makes sense to add some documentation for this! :+1: 

I've just filed #342 to add some additional documentation for the concurrency limit and how to possibly avoid hitting this limit. Let me know what you think about this :+1:

PR #336, opened 2019-02-21
Warning about concurency on dev machines

woytam:
Warning explicitly about `LimitConcurrentRequestsMiddleware` in `Server` on development machines


clue:
@woytam Thank you for filing this PR, I agree that it makes sense to add some documentation for this! :+1: 

I've just filed #342 to add some additional documentation for the concurrency limit and how to possibly avoid hitting this limit. Let me know what you think about this :+1:

PR #364, opened 2020-06-30
Run tests on PHPUnit 9 and clean up test suite

SimonFrings:


tests/TestCase.php
```diff
@@ -63,8 +63,46 @@ protected function expectCallableConsecutive($numberOfCalls, array $with)
 
     protected function createCallableMock()
     {
-        return $this
-            ->getMockBuilder('React\Tests\Http\CallableStub')
-            ->getMock();
+        return $this->getMockBuilder('stdClass')->setMethods(array('__invoke'))->getMock();
```
clue, line 66: LGTM, but `CallableStub` should be removed in this case :+1:

PR #368, opened 2020-07-07
Add HTTP client implementation (import clue/reactphp-buzz v2.9.0)

### LINKED ISSUE #148: Merge HttpClient component into this component
Here's the current situation:
* This Http component: Event-driven, streaming plaintext HTTP and secure HTTPS server for ReactPHP. 
* HttpClient component: Asynchronous HTTP client library. 

This is not exactly ideal, they share quite a bit of common code.

Also, I believe this situation may be a bit confusing for consumers of this package. In particular our Datagram component provides both client and server side for datagram sockets (UDP).

As such, I'd vote for merging these two components into a single Http component.

clue:
This changeset adds an HTTP client implementation by importing clue/reactphp-buzz v2.9.0 and the underlying react/http-client v0.5.10. This client implementation uses a Promise-based API using common PSR-7 messages just like the existing server implementation, so they share common interfaces and concepts.

```php
$browser = new React\Http\Browser($loop);
$browser->get($url)->then(function (Psr\Http\Message\ResponseInterface $response) {
    echo $response->getBody();
});
```

The code has been imported as-is, with only minor changes to the namespace from `Clue\React\Buzz` to `React\Http` and we otherwise leave all the existing APIs unchanged. Deprecated client APIs have been dropped for this major release. As such, this is a pure feature addition with no BC breaks. Upgrading from clue/reactphp-buzz 2.9.0 to the upcoming react/http v1.0.0 should be a matter of updating some namespace references only:

```php
// old
$browser = new Clue\React\Buzz\Browser($loop);

// new
$browser = new React\Http\Browser($loop);
```

By promoting my existing HTTP client implementation clue/reactphp-buzz, we can now provide better support for this component and take advantage of a shared code base. Among others, a follow-up will eliminate a number of minor code duplicates and improve code reuse for specific HTTP edge cases. Upcoming changes will focus on some exciting and long-awaited features, but more on that in follow-up PRs.

See https://github.com/clue/reactphp-buzz for original repo. I'm planning to soft-deprecate this repository in the future and will redirect users to this HTTP component instead.

Resolves #148


legionth:
Nice 👍

Issue #376, opened 2020-07-12
HTTP client: Support HTTP upgrades (WebSockets)


clue:
This library should provide support for HTTP upgrades (WebSocket etc.) using the `Connection: upgrade` and `Upgrade: websocket` headers.

This is already supported on the server-side, so adding support for this on the client-side should be reasonable. This will likely require a dedicated API as the existing streaming APIs only expose a `ReadableStreamInterface` and the upgraded stream would have to be a `DuplexStreamInterface`.

bosunski:
The good news is that I was able to have a working implementation of this during the weekend 😀. I'm currently testing and hope to get a PR in by tomorrow.

mpociot:
@clue Are there any plans to add this feature?
I'd love to add the ability to forward websocket connections to Expose, but those requests currently get closed immediately

PR #382, opened 2020-07-27
[WIP] Expose DuplexStreamInterface for connections that requires Upgrades

bosunski:
Based on the Idea of #376 by @clue 

Emits an `upgrade` event when:

-  `Connection` header is present in the response headers and values has `upgrade`
- Response status code is `101` 

Event emitted exposes a `React\Stream\DuplexStreamInterface`, a `Psr\Http\Message\ResponseInterface` and `React\Http\Client\Request`.

## Summary of changes
- Adds custom header parser to parse response data before data is passed down to the main `data` event parser.
- Adds detection of Upgraded responses
- Updated `React\Http\Io\Sender` and `React\Http\Io\Transaction` to be configurable for Upgrade connections

Example Usage:

```php
$request = $client->request("GET", 'https://echo.websocket.org:443', ['Connection' => 'Upgrade', 'Upgrade' => 'websocket']);

$request->on('upgrade', function (DuplexStreamInterface $connection, $response, $request) {
    echo "Request Upgraded";
});
```
Websocket Client Usage:
```php
$promise = $browser->withOptions(['upgrade' => true])->get('https://echo.websocket.org:443', ['Connection' => 'Upgrade', 'Upgrade' => 'websocket']);

$response->then(function (UpgradedResponse $response) use ($loop) {
    $ws = new \Ratchet\ClientWebSocket($response->getConnection(), $response->getResponse(), $response->getRequest());
});
```

Example Usage in `clue/reactphp-docker`

```php
$this->browser->withOptions(array('streaming' => true, 'upgrade' => true))->post(
            $this->uri->expand(
                '/exec/{exec}/start',
                array(
                    'exec' => $exec
                )
            ),
            array(
                'Content-Type' => 'application/json',
                'Connection' => 'Upgrade',
                'Upgrade' => 'tcp',
            ),
            $this->json(array(
                'Tty' => !!$tty
            ))
  );
```


bosunski:
Still working some failing tests, you can please help review while I fix the tests.
Thanks!
cc @clue @WyriHaximus

src/Client/Request.php
```diff
@@ -138,7 +146,16 @@ public function handleData($data)
         // buffer until double CRLF (or double LF for compatibility with legacy servers)
         if (false !== strpos($this->buffer, "\r\n\r\n") || false !== strpos($this->buffer, "\n\n")) {
             try {
-                list($response, $bodyChunk) = $this->parseResponse($this->buffer);
+                $psrResponse = gPsr\parse_response($this->buffer);
+
+                if ($this->responseIsAnUpgradeResponse($psrResponse)) {
+                    $this->stream->removeListener('data', array($this, 'handleData'));
+
+                    $this->emit('upgrade', array($this->stream, $psrResponse, $this));
```
clue, line 154: Not sure about the whole `upgrade` event. How does this interfere with the existing request/response semantics?

I like how this is exposed on the server side via https://github.com/reactphp/http#streaming-outgoing-response, perhaps we can find a similar solution here?

clue (commented):
@bosunski Thanks for looking into this, would love to get this feature in!

It looks like this is still a work in progress, but I like how this doesn't seem to affect a lot of code at the moment. I've added some minor remarks below.

Let us know if there's anything we can help with. Keep it up! 💪

src/Browser.php
```diff
@@ -707,7 +708,7 @@ public function withResponseBuffer($maximumSize)
      * @see self::withFollowRedirects()
      * @see self::withRejectErrorResponse()
      */
-    private function withOptions(array $options)
+    public function withOptions(array $options)
```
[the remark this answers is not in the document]
bosunski: @clue I didn't know the method was deprecated.

But my thought behind it was to let the consumer using `Browser` to be able to choose to watch out for the upgrade event or not. As by default, `Browser` will not. This means that, if an upgrade did happen, I need to configure browser to capture it like `$browser->withOptions(['upgrade' => true])` (default is `false` in `Transaction.php`). 

This, however, won't be necessary if the default mode should be to capture the upgrade. That was the reason I made the method public (also seeing it was public in `clue/reactphp-buzz`)

src/Browser.php
```diff
@@ -720,7 +721,7 @@ private function withOptions(array $options)
      * @param string                         $url
      * @param array                          $headers
      * @param string|ReadableStreamInterface $body
-     * @return PromiseInterface<ResponseInterface,Exception>
+     * @return PromiseInterface<ResponseInterface,Exception, ConnectionInterface>
```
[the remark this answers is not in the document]
bosunski: I added it in the case, where an `upgrade` did happen, the promise will resolve with a `ConnectionInterface` rather than an `Exception` or a `ResponseInterface`. Still not sure if this is the best way to express that.

src/Client/Request.php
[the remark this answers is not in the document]
bosunski: Not sure if this is the right `Request` object to expose, but it's kind of a good idea to expose `a` request object because there could be cases where you want to verify some things about the request **after** upgrade and then close the connection if that verification fails. That's my thought about it. But sure, I might be wrong here.

I actually read a piece of Node docs to see how `upgrade` is being handled by the consumer side. I followed that in adding the upgrade event also. [Here](https://nodejs.org/api/http.html#http_event_upgrade) is a sample piece of how node handles the upgrade.

Sure, If there's a better you think this could be or should be done, I'll be willing to make it work.

bosunski: I actually followed a Node piece I read [here](https://nodejs.org/api/http.html#http_event_upgrade) as I also mentioned above.

I will check out the sample you sent, I kind of like it, knowing that a `ThroughStream` also implements` DuplexStreamInterface`.

bosunski:
Hello @clue , are there anything that can be done to move this PR forward since this is still a desired feature?

src/Client/UpgradedResponse.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @return RequestInterface
+     */
+    public function getRequest()
+    {
+        return $this->request;
+    }
+}
```
clue, line 56: nit: missing EOL

src/Client/UpgradedResponse.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace React\Http\Client;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use React\Stream\DuplexStreamInterface;
+
+class UpgradedResponse
```
clue, line 9: This adds a new class as part of our public API in the internal `React\Http\Client` namespace.

src/Browser.php
```diff
@@ -770,7 +771,7 @@ private function withOptions(array $options)
      * @param string                         $url
      * @param array                          $headers
      * @param string|ReadableStreamInterface $body
-     * @return PromiseInterface<ResponseInterface,\Exception>
+     * @return PromiseInterface<ResponseInterface,\Exception,ConnectionInterface>
```
clue, line 774: What is this supposed to do? The `PromiseInterface` does not currently nor the foreseeable future have a third template type.

clue:
> Hello @clue , are there anything that can be done to move this PR forward since this is still a desired feature?

@bosunski Thank you for the friendly reminder! I agree this feature would be nice to have, but in all honestly I don't see the currently suggested API as a viable alternative. We're moving towards a more type-safe future (see also https://github.com/orgs/reactphp/discussions/472) and this PR currently seems to add APIs that can not be covered by the existing type system. I think we're on the same page that the underlying feature request is still interesting, but I wonder what a decent API could look like?

bosunski:
Thanks for your response @clue

This PR is already 2 years old 😅 and I also agree that this PR needs a reimplementation to be consistent with the changes that have happened within ReactPHP itself and also to align with the current goals.

I'm still very interested in getting this out of the way. 

Maybe we can start by talking about how the upgrade API will be used from a developers' point of view.

clue:
@bosunski Agreed, let's continue the discussion in the feature ticket #376 and close this PR. May I ask you to post a quick gist in that ticket so others can follow what's been decided so far? Thank you!

PR #399, opened 2021-01-06
Fix a typo, a mixed up param name and add forgotten namespace backslash

pavog:
Hello, I noticed a few minor problems in the browser that I would like to fix with this pr.

I renamed the "$contents" parameter in requestStreaming() to "$body" to match the other functions where it is called "$body".

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -15,7 +15,7 @@ use React\Stream\ReadableStreamInterface;
 use InvalidArgumentException;
 
 /**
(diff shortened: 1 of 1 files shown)


WyriHaximus (approved):

clue (approved):
@pavog Thanks for looking into this, the suggested changes LGTM! :shipit: 

The change refs https://github.com/reactphp/http/issues/401, but I suppose it makes more sense to address this individually :+1:

jsor (approved):

PR #400, opened 2021-01-06
Update example 63 to support v1.0.0+

pavog:
Add LoopInterface as first constructor argument to Server and change Server to accept variadic middleware handlers instead of array.


WyriHaximus (approved):

clue (approved):
@pavog Thanks for the update, changes LGTM! :shipit: 

This does indeed appear to be a small oversight from https://github.com/reactphp/http/pull/361 / https://github.com/reactphp/http/pull/362

jsor (approved):

PR #393, opened 2020-09-02
Update PHPUnit configuration schema for PHPUnit 9.3 and replace deprecated at() Mocks

SimonFrings:
PHPUnit 9.3 released a new schema for the `phpunit.xml` configuration file. I had to migrate the file to the new format in order to avoid the warning. PHPUnit Versions older than 9.3 have to use the `phpunit.xml.legacy` configuration file, because the new format is unknown for them.
I also had to update  some tests, because the `at()` matcher deprecated in PHPUnit 9.3 and will be removed in PHPUnit 10.

Together with the changes from https://github.com/reactphp/http/pull/391 by @clue it's possible to run this code with PHP 8 without any warnings in your output 🎉
 ```bash
$ docker run -it --rm -v `pwd`:/data --workdir=/data php:8.0.0beta2-cli php vendor/bin/phpunit
PHPUnit 9.3.8 by Sebastian Bergmann and contributors.

...............................................................  63 / 673 (  9%)
............................................................... 126 / 673 ( 18%)
............................................................... 189 / 673 ( 28%)
............................................................... 252 / 673 ( 37%)
............................................................... 315 / 673 ( 46%)
............................................................... 378 / 673 ( 56%)
............................................................... 441 / 673 ( 65%)
............................................................... 504 / 673 ( 74%)
............................................................... 567 / 673 ( 84%)
............................................................... 630 / 673 ( 93%)
...........................................                     673 / 673 (100%)

Time: 00:16.651, Memory: 20.00 MB

OK (673 tests, 1604 assertions)
```

For further details concerning this pull request look into [graphp/graphviz#46](https://github.com/graphp/graphviz/pull/46).
This pull request builds on top of [reactphp/http#364](https://github.com/reactphp/http/pull/364).

PR #402, opened 2021-02-03
Use GitHub actions for continuous integration (CI)

SimonFrings:
Bye bye Travis CI, you've served us well.
@clue and I are currently working on switching from Travis to GitHub actions in all ReactPHP repositories.

I've built a little workaround for some tests for legacy PHP. When using recursion there is always the risk of reaching a full stack which results in PHP failing hard. Therefore the `max_nesting_level` setting from Xdebug (to prevent reaching this point) was set to 100 for PHP Versions < 5.4 which resulted in failing test for legacy PHP. In 2015 this setting was changed to 256 for PHP Versions > 5.4. For more details look into [Xdebug#1100](https://bugs.xdebug.org/bug_view_page.php?bug_id=00001100) and reactphp/dns#170.

This code supports running tests on PHP 8.

Builds on top of #393 and reactphp/reactphp#442.


clue (approved):
@SimonFrings Thanks for taking over the update, changes LGTM! :shipit:

WyriHaximus (approved):

jsor (approved):

PR #419, opened 2021-07-27
Simplify usage by supporting new Socket API without nullable loop arguments

clue:
This changeset simplifies usage by supporting the new Socket API without nullable loop arguments.

```php
// old
$http->listen(new React\Socket\Server(8080));

// new
$http->listen(new React\Socket\SocketServer('127.0.0.1:8080'));
```

Note that this doesn't affect the API of this package at all, but does use the improved API of the referenced Socket component. Existing code continues to work as is.

Together with #418, #417 and #410, this means we can now fully rely on the [default loop](https://github.com/reactphp/event-loop#loop) and no longer need to skip any arguments with `null` values. Additionally, this now consistently uses the `HttpServer` and `SocketServer` class names to avoid class name collisions and ambiguities.

Builds on top of https://github.com/reactphp/socket/pull/264 and https://github.com/reactphp/socket/pull/263

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ $http = new React\Http\HttpServer(function (Psr\Http\Message\ServerRequestInterf
     );
 });
 
-$socket = new React\Socket\Server(8080);
+$socket = new React\Socket\SocketServer('127.0.0.1:8080');
 $http->listen($socket);
 ```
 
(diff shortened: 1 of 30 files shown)


clue:
Rebased now that #418 has been merged and Socket v1.9.0 has been released :shipit:

WyriHaximus (approved):

PR #432, opened 2021-11-07
Expose all status code constants via `Response` class

clue:
With this changeset, we now expose all status code constants via `Response` class. This is done by implementing the [PSR-7 Message Util `StatusCodeInterface`](https://github.com/php-fig/http-message-util/blob/master/src/StatusCodeInterface.php) which means that most common HTTP status codes are now available as class constants with the `STATUS_*` prefix. For instance, the `200 OK` and `404 Not Found` status codes can used as `Response::STATUS_OK` and `Response::STATUS_NOT_FOUND` respectively.

Using status code constants is entirely optional and using status codes as integers remains supported of course, but we now use these constants throughout our documentation and examples to make them more readable. Accordingly, this does not affect backwards compatibility.

This builds on top of #406 (thank you @WyriHaximus!), but instead of exposing a new interface, this changeset simply ensures the existing `Response` class implements this very interface. This makes it much easier to use these constants in application code because no separate import is required.

Supersedes / closes #406

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ require __DIR__ . '/vendor/autoload.php';
 
 $http = new React\Http\HttpServer(function (Psr\Http\Message\ServerRequestInterface $request) {
     return new React\Http\Message\Response(
-        200,
+        React\Http\Message\Response::STATUS_OK,
         array(
             'Content-Type' => 'text/plain'
         ),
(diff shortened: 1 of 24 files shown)


clue:
Rebased on `master` to resolve merge conflict now that #431 is in :shipit:

WyriHaximus (approved):

PR #438, opened 2022-01-23
Update test suite to use default loop

clue:
Builds on top of #410, #417, #418, #419, https://github.com/clue/reactphp-block/pull/60, https://github.com/reactphp/socket/pull/283 and others


WyriHaximus (approved):
Can't wait to use `react/async` for the awaits

PR #430, opened 2021-11-05
Improve assigning `Content-Length` for `304 Not Modified` response

clue:
This changeset ensures we preserve any explicit `Content-Length` response headers and do not assign a `Content-Length: 0` by default for a `304 Not Modified` response.

Builds on top of #429, #188 and #156
Refs https://github.com/tornadoweb/tornado/issues/1736 and https://github.com/ninenines/cowboy/issues/1153

PR #444, opened 2022-03-03
Allow Explicit content length on HEAD requests

mrsimonbennett:
Hey guys, I am in the process of building a server to site infront of S3

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13

"The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the **case of the HEAD method**, the **size of the entity-body that would have been sent had the request been a GET.**"


clue (commented):
@mrsimonbennett Thank you for filing this PR, changes LGTM!

I've reproduced this locally and can can confirm this changeset is indeed needed to give more control over the `Content-Length` response header for `HEAD` requests. This is somewhat related to #429 and #430 merged only recently.

May I ask you to squash your changes into a single commit? :+1:

mrsimonbennett:
@clue done :)

clue (approved):
@mrsimonbennett Thanks for the update, changes LGTM! :shipit: Keep it up :+1:

WyriHaximus (approved):

PR #472, opened 2022-09-17
Set up PHPStan on GitHub Actions

nhedger:
This PR sets up PHPStan to run on GitHub Actions, as discussed in [discussions#469](https://github.com/orgs/reactphp/discussions/469).

### Overview

- [x] Sets up PHPStan to run on GitHub Actions on PHP 8.1 only
- [x] Configures PHPStan to run the analysis on the `examples`, `src` and `tests` folders
- [x] Generates the [baseline](https://phpstan.org/user-guide/baseline) so that static analysis passes immediately 

### Baseline

Because this PR aims to set up PHPStan and not address the errors it reports, I've generated a baseline to make the pipeline succeed. We'll then be able to incrementally fix the problems in future PRs.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,11 @@
+includes:
+  - phpstan-baseline.neon
+
+parameters:
+  level: 9
+  paths:
+    - examples
+    - src
+    - tests
+  excludePaths:
+    - src/Server.php
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,3 +47,23 @@ jobs:
       - run: composer self-update --2.2 # downgrade Composer for HHVM
       - run: hhvm $(which composer) install
       - run: hhvm vendor/bin/phpunit
+
+  PHPStan:
+    name: PHPStan
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php: [ 8.1 ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          tools: phpstan
+      - name: Install Composer dependencies
+        run: composer install
+      - name: Run static analysis
+        run: phpstan analyse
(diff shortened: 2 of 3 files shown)


phpstan.neon
```diff
@@ -0,0 +1,11 @@
...
+  - phpstan-baseline.neon
+
+parameters:
+  level: 9
+  paths:
+    - examples
+    - src
+    - tests
+  excludePaths:
+    - src/Server.php
```
nhedger, line 11: I had to exclude this file because otherwise I could not generate the baseline.

This is due to the `Server` class attempting to extend the final `HttpServer` class

nhedger:
The generated baseline produces different results on different operating systems.

The pipeline succeeds when running on macOS but fails when running on Ubuntu (in the GitHub Actions runner at least).

```
 ------ ----------------------------------------------------------- 
  Line   examples/92-client-benchmark-upload.php                    
 ------ ----------------------------------------------------------- 
  109    Call to an undefined method ChunkRepeater::getPosition().  
  116    Call to an undefined method ChunkRepeater::getPosition().  
  116    Call to an undefined method ChunkRepeater::getPosition().  
 ------ ----------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   examples/99-server-benchmark-download.php                              
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^Call to an undefined method                    
         ChunkRepeater\:\:getSize\(\)\.$# in path                               
         /home/runner/work/http/http/examples/99-server-benchmark-download.php  
         was not matched in reported errors.                                    
 ------ ----------------------------------------------------------------------- 
```

It seems to complain about a method not being defined although the method is defined. I cannot find a reason for this currently.

### Possible solution

As a temporary solution, I could probably ignore the problematic files, which are examples anyway.

clue:
For the reference: It looks like this wasn't closed intentionally (as discussed in https://github.com/reactphp/async/pull/64). We've recently started applying PHPStan on max level into our projects with https://github.com/reactphp/async/pull/76 and we'll go ahead with this and apply PHPStan on max level to all our projects in the future.

Thank you very much for your contribution and sparking this discussion! :+1:

PR #370, opened 2020-07-09
Move `Response` class to `React\Http\Message\Response` and `ServerRequest` class to `React\Http\Message\ServerRequest`

clue:
This changeset moves the `Response` class to `React\Http\Message\Response` and the `ServerRequest` class to `React\Http\Message\ServerRequest`. The `Response` class is used throughout our examples and user code, the `ServerRequest` class used to be internal only and is now exposed (testing client web applications).

```php
// old
$response = new React\Http\Response(200, [], 'Hello!');

// new
$response = new React\Http\Message\Response(200, [], 'Hello!');
```

Builds on top of #361 and #368

Issue #331, opened 2018-11-13
Why ringcentral/psr7 and not guzzle/psr7


phpbg:
Hi,

this is just a question, not an issue.

Why do you require `ringcentral/psr7` from https://github.com/ringcentral/psr7 and not `guzzle/psr7` from https://github.com/guzzle/psr7 ?

I see ringcentral is a fork of guzzle, but it doesn't explain why, and what's changed.

Wouldn't guzzle be a better choice?

jsor:
See #319

phpbg:
Mmmhh ok, sorry for the duplicate.
Don't you want to update the minimum requirements for `react/http` ?
5.3 is quite obsolete now...
Regards

clue:
"Why guzzle/psr7 and not ringcentral/psr7?" :-)

I would argue that it doesn't really make much of a difference which PSR-7 implementation we're using. For the most part, this is an internal implementation detail that is not in any way exposed to the outside and our documentation explicitly says we're using PSR-7 interfaces only.

While I can see where you're coming from, I don't think discussing PHP version requirements adds much to this specific discussion.

Additionally, there are plans to drop RingCentral in a future version. Partly because it adds a lot of stuff that we don't really use here (sync file handling etc.) and partly because it's missing some stuff that we do need here (async implementations etc.).

I believe this has been answered, so I'm closing this for now. Please come back with more details if this problem persists and we can reopen this :+1:

PR #480, opened 2022-11-23
Add `Request` class to represent outgoing HTTP request message

clue:
This changeset adds a `Request` class to represent an outgoing HTTP request message. This class implements the [PSR-7 `RequestInterface`](https://www.php-fig.org/psr/psr-7/#32-psrhttpmessagerequestinterface) which extends the [PSR-7 `MessageInterface`](https://www.php-fig.org/psr/psr-7/#31-psrhttpmessagemessageinterface). It is mostly used internally to represent each outgoing HTTP request message for the HTTP client implementation. Likewise, you can also use this class with other HTTP client implementations and for tests.

This class complements our other PSR-7 implementations (#370 and others) and is one step further towards moving away from the RingCentral implementations as discussed in #331/#437. On top of this, this PR is done in preparation for HTTP keep-alive support as discussed in #468/#39. Once this PR is merged, I'll file the next follow-up PRs to reuse this class for our outgoing request messages much like already done for the incoming response messages as per #389.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ multiple concurrent HTTP requests without blocking.
             * [json()](#json)
             * [plaintext()](#plaintext)
             * [xml()](#xml)
+        * [Request](#request-1)
         * [ServerRequest](#serverrequest)
         * [ResponseException](#responseexception)
     * [React\Http\Middleware](#reacthttpmiddleware)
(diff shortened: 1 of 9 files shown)


SimonFrings (approved):
Looks good to me :+1:

WyriHaximus (approved):

PR #481, opened 2022-11-29
Internal refactoring to remove internal `RequestData` and rename internal `Request` to `ClientRequestStream`

clue:
This changeset refactors some internal logic to remove the internal `RequestData` and reuse the request object instead (#480). Additionally, it renames the internal only `React\Http\Client\Request` class to `React\Http\Io\ClientRequestStream` to avoid confusion.

This does not affect the public `React\Http\Message\Request` class or any other public APIs, so this should be safe to apply. The test suite confirms this has 100% code coverage and does not otherwise affect our APIs.

Builds on top of #480 and #389
Done in preparation for #468


SimonFrings (approved):
:+1:

WyriHaximus (approved):

PR #484, opened 2023-01-18
Consistently close HTTP client connections when response stream closes

clue:
This changeset improves how the HTTP client closes outgoing TCP/IP connections when the response stream closes. This should not affect "normal" operation of well-behaving peers, but would be relatively easy to trigger depending on which HTTP headers the request and response specified. Previously, connections may have been kept in an idle state instead of properly closing the connection and freeing underlying system resources.

Applying these changes required a significant refactoring of the existing logic first and also happens to be a prerequisite for upcoming HTTP keep-alive handling that will be taken care of in a follow-up PR. This does not otherwise affect the public `React\Http\Browser` class or any other public APIs, so this should be safe to apply. The test suite confirms this has 100% code coverage and does not otherwise affect our public APIs.

Builds on top of #481 and #480
Done in preparation for #468 (HTTP keep-alive)
Also refs #376 (future support for HTTP upgrades)

### DIFF
--- a/src/Io/ClientRequestStream.php
+++ b/src/Io/ClientRequestStream.php
@@ -4,6 +4,8 @@ namespace React\Http\Io;
 
 use Evenement\EventEmitter;
 use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use React\Http\Message\Response;
 use React\Promise;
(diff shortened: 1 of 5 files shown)


WyriHaximus (approved):

SimonFrings (approved):
Nice :+1:

PR #485, opened 2023-01-18
Removed obsolete RingCentral/Psr7 dependency

vanodevium:
[ringcentral/psr7](https://packagist.org/packages/ringcentral/psr7) hasn't been updated in over 5 years.

Need to keep up with the times.

None of the existing PSR7 implementations work on PHP >=5.3

So I collect code from the latest versions of [guzzle/psr7](https://packagist.org/packages/guzzlehttp/psr7) and [nyholm/psr7](https://packagist.org/packages/nyholm/psr7) and made own local realization which can work on PHP 5.3 and upper.

This code passes all tests.

if you're interested, I can continue work on modifying any existing implementation enough to create your own React/Psr7 package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,7 @@
         "react/event-loop": "^1.2",
         "react/promise": "^3 || ^2.3 || ^1.2.1",
         "react/socket": "^1.12",
-        "react/stream": "^1.2",
-        "ringcentral/psr7": "^1.2"
+        "react/stream": "^1.2"
     },
     "require-dev": {
         "clue/http-proxy-react": "^1.8",
--- a/examples/11-client-http-proxy.php
+++ b/examples/11-client-http-proxy.php
@@ -25,7 +25,7 @@ $browser = new Browser($connector);
 
 // demo fetching HTTP headers (or bail out otherwise)
 $browser->get('https://www.google.com/')->then(function (ResponseInterface $response) {
-    echo RingCentral\Psr7\str($response);
+    echo React\Http\Psr7\str($response);
 }, function (Exception $e) {
     echo 'Error: ' . $e->getMessage() . PHP_EOL;
 });
--- a/examples/12-client-socks-proxy.php
+++ b/examples/12-client-socks-proxy.php
@@ -25,7 +25,7 @@ $browser = new Browser($connector);
 
 // demo fetching HTTP headers (or bail out otherwise)
 $browser->get('https://www.google.com/')->then(function (ResponseInterface $response) {
-    echo RingCentral\Psr7\str($response);
+    echo React\Http\Psr7\str($response);
 }, function (Exception $e) {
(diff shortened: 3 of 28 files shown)


SimonFrings:
@vanodevium  thanks for bringing this up, always happy about contributions :+1:

I agree that it makes sense to add this as a part of ReactPHP at some point and I really appreciate the effort you put into this, but I don't think this is the best way to do it for various reasons. Using existing code from other projects running under the MIT license need to be addressed, otherwise we would violate this license. 

Additionally I think we should also add only the necessary classes and functions that we use in our APIs, we don't want to bloat the project with unneeded functionality (which is also marked internal, so if we don't use it, who will).

Apart from that we also need to add tests for the new added classes. Updating the existing tests to use the newer classes is correct in itself, but it doesn't cover every new function added in this PR.

We're aware that we can't use [guzzle/psr7](https://packagist.org/packages/guzzlehttp/psr7) and [nyholm/psr7](https://packagist.org/packages/nyholm/psr7) because of old PHP, but this won't be the case once ReactPHP v3 is out which gets rid of "old" PHP. Maybe it makes sense to wait for v3 to come out, which will remove a lot of the old stuff and then focus on implementing this as a part of ReactPHP.

What do you think?

vanodevium:
I'm glad that you like it.

With this code, I just showed that, in general, you can get rid of the old dependency and make your own internal implementation.

If you are interested, just let me know and I will write an implementation with my own hands that will not violate any licenses.

Regarding the tests, I do not quite agree: I only changed the base classes that the project classes inherit, but they are already completely covered by tests and there is no new functionality.

Of course, if it is possible to wait for a new version of your project, then it is better to wait.

Thank you for paying attention to my work. Always glad to take part and contribute to the development of great projects!

vanodevium:
one more thing

I absolutely fully understand what I did in this code, because ringcentral/psr7 is actually a very old fork of the guzzle/psr7

Technically, I just updated it :) with NO breaking changes

And last thing:
maybe it will be nice to integrate **psr7-integration-tests**

SimonFrings:
> If you are interested, just let me know and I will write an implementation with my own hands that will not violate any licenses.

I like your motivation :+1: @clue told me about his vision on how we can integrate this into ReactPHP's HTTP component, but I think we have to work out the details first and come up with a more solid plan for this. As I said above, we're currently busy with the ReactPHP v3 development, which I think will be a great foundation for a future PSR-7 implementation.

Better support for PSRs is also one of our goals for ReactPHP v3.x as described in our [plans for the upcoming major version](https://github.com/orgs/reactphp/discussions/481), not sure how much of it will make it in time for the v3 release tho.

SimonFrings:
The RingCentral dependency has now officially been removed in #522 and we're now using our own PSR-7 implementation instead (see #518, #519, #520, #521)

This means there's nothing more to do here and I'll close this pull request. Thank you @vanodevium for working on this! :heart:

PR #502, opened 2023-07-07
Update tests to remove defunct httpbin.org

clue:
Bye bye httpbin.org, you've served us well.

The service has been down sporadically for a few months already and there doesn't appear to be a clear path forward (see https://github.com/postmanlabs/httpbin/issues/702 and others). I've replaced most occurrences with httpbingo.org as a drop-in replacement where possible.

We've introduced this a while back via https://github.com/reactphp/http-client/pull/109 before merging the HTTP client with #368 (see also https://github.com/reactphp/socket/issues/105 and #244) and haven't touched this ever since. These tests don't provide a lot of value nowadays given that this is a rather high-level test and the underlying logic is already covered by our unit test suite. Accordingly, removing these tests has no impact on our code coverage.

See also #480, #481 and others, as we'll likely eliminate the legacy `Client` class in the future


SimonFrings (approved):
Makes sense, no more failing tests because of this :+1:

WyriHaximus (approved):

Issue #504, opened 2023-08-15
Laravel + websockets: Unhandled promise rejection with TypeError


ryanbaldisco:
I recently updated my laravel project from (laravel/framework) 10.0 to 10.18, as well as all of my packages. Everything was updated smoothly without errors or warnings. However, when I tried to run the websockets server `php artisan websockets:serve` I got this error below.

```
Unhandled promise rejection with TypeError: React\Http\Io\ClientRequestStream::closeError(): Argument #1 ($error) must be of type Exception, Error given, called in /path/to/laravel/vendor/react/promise/src/Internal/RejectedPromise.php on line 73 in /path/to/laravel/vendor/react/http/src/Io/ClientRequestStream.php:238
```

I tried debugging for hours, but had no luck.
Please, can anyone help me with this? or at least point me to the right direction.

```
PHP version 8.2.5
Composer version 2.5.8
```

Stack trace:
```
#0 /path/to/laravel/vendor/react/promise/src/Internal/RejectedPromise.php(73): React\Http\Io\ClientRequestStream->closeError(Object(Error))
#1 /path/to/laravel/vendor/react/promise/src/Promise.php(43): React\Promise\Internal\RejectedPromise->then(Object(Closure), Array)
#2 /path/to/laravel/vendor/react/http/src/Io/ClientRequestStream.php(67): React\Promise\Promise->then(Object(Closure), Array)
#3 /path/to/laravel/vendor/react/http/src/Io/ClientRequestStream.php(122): React\Http\Io\ClientRequestStream->writeHead()
#4 /path/to/laravel/vendor/react/http/src/Io/ClientRequestStream.php(135): React\Http\Io\ClientRequestStream->write('{"app_id":"6087...')
#5 /path/to/laravel/vendor/react/http/src/Io/Sender.php(152): React\Http\Io\ClientRequestStream->end('{"app_id":"6087...')
#6 /path/to/laravel/vendor/react/http/src/Io/Transaction.php(146): React\Http\Io\Sender->send(Object(React\Http\Message\Request))
#7 /path/to/laravel/vendor/react/http/src/Io/Transaction.php(83): React\Http\Io\Transaction->next(Object(React\Http\Message\Request), Object(React\Promise\Deferred), Object(React\Http\Io\ClientRequestState))
#8 /path/to/laravel/vendor/react/http/src/Browser.php(853): React\Http\Io\Transaction->send(Object(React\Http\Message\Request))
[...]

PR #506, opened 2023-10-11
fix: #504 laravel + websockets: Unhandled promise rejection with TypeError

SocolaDaiCa:
#504

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0",
+        "php": ">=7.0",
         "evenement/evenement": "^3.0 || ^2.0 || ^1.0",
         "fig/http-message-util": "^1.1",
         "psr/http-message": "^1.0",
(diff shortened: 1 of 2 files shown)


SimonFrings:
@SocolaDaiCa thanks for trying to fix the issue reported in #504 :+1:

I just gave an explanation in #504 why this isn't really a bug in our HTTP component and should be reported downstream in https://github.com/beyondcode/laravel-websockets instead.

I can totally understand the reason behind this pull request, as it makes sense that we should support  Throwables instead of just exceptions. Raising the PHP version from PHP 5.3 to PHP 7 is at least a BC break and I don't think we want to go this step for the current v1.

We're currently working on ReactPHP v3 which will raise the minimal supported PHP version to PHP 7.1 anyway, so I think we might want to take a look at this over there. We also have plans to improve the error reporting for v3, so I think we will pick this up in the upcoming weeks/months. The v3 roadmap ticket for HTTP is still outstanding but will drop soon. 

With all that said, I'll close this ticket for now and we'll have a look at this as we move forward with ReactPHP v3 :rocket:

PR #518, opened 2024-03-13
Refactor `Response` class to build on top of new PSR-7 implementation

clue:
This changeset refactors the `Response` class to build on top of a new PSR-7 implementation. This brings us one step closer to eventually replace the dated RingCentral implementation (#331) and eventually support PSR-7 v2 (#513). This is a purely internal change that comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply.

As a side effect, this appears to show a minor, but noticeable, performance improvement during my benchmarks:

```bash
$ php examples/99-server-benchmark-download.php 127.0.0.1:8080
$ docker run -it --rm --net=host jordi/ab -n100000 -c100 -k http://localhost:8080/

# old: 48428 requests per second
# new: 51288 requests per second
```

(Also posted in https://twitter.com/another_clue/status/1767924579908174213)

Once merged, I'll file follow-up PRs to refactor the `Request` and `ServerRequest` classes (#480 and #170) accordingly. If you enjoy this change and want to help us continue to ship more improvements, consider supporting this project, for example by [becoming a sponsor](https://github.com/sponsors/reactphp) ❤️

Builds on top of #432, #370, #480 and others

### DIFF
--- a/README.md
+++ b/README.md
@@ -2448,8 +2448,7 @@ constants with the `STATUS_*` prefix. For instance, the `200 OK` and
 `404 Not Found` status codes can used as `Response::STATUS_OK` and
 `Response::STATUS_NOT_FOUND` respectively.
 
-> Internally, this implementation builds on top of an existing incoming
-  response message and only adds required streaming support. This base class is
+> Internally, this implementation builds on top of a base class which is
   considered an implementation detail that may change in the future.
 
 ##### html()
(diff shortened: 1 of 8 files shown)


SimonFrings (approved):
Looks good to me :+1:

WyriHaximus (approved):
:shipit:

PR #519, opened 2024-03-15
Refactor `Request` and `ServerRequest` classes to build on top of new PSR-7 implementation

clue:
This changeset refactors the `Request` and `ServerRequest` classes to build on top of a new PSR-7 implementation. This brings us one step closer to eventually replace the dated RingCentral implementation (#331) and eventually support PSR-7 v2 (#513). This is a purely internal change that comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply.

This builds on top of the recent changes for the `Response` class (#518), but this time doesn't show a noticeable impact on performance during my benchmarks. As a consequence, I consider this mostly an internal refactoring only.

Once merged, I'll file follow-up PRs to refactor the `Uri` and internal parsing classes accordingly. If you enjoy this change and want to help us continue to ship more improvements, consider supporting this project, for example by [becoming a sponsor](https://github.com/sponsors/reactphp) ❤️

Builds on top of #518, #480, #170 and others

### DIFF
--- a/README.md
+++ b/README.md
@@ -2642,8 +2642,7 @@ This is mostly used internally to represent each outgoing HTTP request
 message for the HTTP client implementation. Likewise, you can also use this
 class with other HTTP client implementations and for tests.
 
-> Internally, this implementation builds on top of an existing outgoing
-  request message and only adds support for streaming. This base class is
+> Internally, this implementation builds on top of a base class which is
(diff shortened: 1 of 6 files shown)


SimonFrings (approved):
Looks good to me, let's ship it :+1: :duck:

WyriHaximus (approved):

PR #520, opened 2024-03-22
Refactor HTTP message parsing to build on top of new PSR-7 implementation

clue:
This changeset refactors the HTTP message parsing logic in the internal `ClientRequestStream` and `RequestHeaderParser` classes to build on top of a new PSR-7 implementation. This brings us one step closer to eventually replace the dated RingCentral implementation (#331) and eventually support PSR-7 v2 (#513). Most notably, this means the `Browser` will now return instances of our `Response` class (#518) instead of using the legacy RingCentral classes.

Other than that, this is a purely internal change that comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply. I've tried to keep changes to a minimum to ease review, but there's definitely some room for improvement in follow-up PRs.

This builds on top of the recent changes for the `Response` and `ServerRequest` classes (#518 and #519), but this time doesn't show a noticeable impact on performance during my benchmarks. As a consequence, I consider this mostly an internal refactoring only.

Once merged, I'll file follow-up PRs to refactor the `Uri` class accordingly. If you enjoy this change and want to help us continue to ship more improvements, consider supporting this project, for example by [becoming a sponsor](https://github.com/sponsors/reactphp) ❤️

Builds on top of #519, #518, #480, #432, #370, #170 and others


SimonFrings (approved):
Went through all changes, looks good so let's get this shipped :+1:

WyriHaximus (approved):
:shipit:

PR #521, opened 2024-03-25
Add new `Uri` class for new PSR-7 implementation

clue:
This changeset adds a new `Uri` class that builds on top of our new PSR-7 implementation. This brings us one step closer to eventually replace the dated RingCentral implementation (#331) and eventually support PSR-7 v2 (#513). This is a mostly an internal change, but also exposes the new `Uri` class as part of our public API. Perhaps most notably, this means the `Request` and `ServerRequest` classes will now return instances of our `Uri` class (#519) instead of using the legacy RingCentral classes. It comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply.

This builds on top of the recent changes for the `Response` and `ServerRequest` classes (#518 and #519), but this similar to #520 doesn't show a noticeable impact on performance during my benchmarks. As a consequence, I consider this mostly a simple feature addition only.

Once merged, I'll file follow-up PRs to remove any leftover references to the old RingCentral implementation accordingly. If you enjoy this change and want to help us continue to ship more improvements, consider supporting this project, for example by [becoming a sponsor](https://github.com/sponsors/reactphp) ❤️

Builds on top of #520, #519, #518, #480, #432, #370, #170 and others

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ multiple concurrent HTTP requests without blocking.
             * [xml()](#xml)
         * [Request](#request-1)
         * [ServerRequest](#serverrequest)
+        * [Uri](#uri)
         * [ResponseException](#responseexception)
     * [React\Http\Middleware](#reacthttpmiddleware)
         * [StreamingRequestMiddleware](#streamingrequestmiddleware)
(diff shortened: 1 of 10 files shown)


SimonFrings (approved):
Nice, let's get this in :+1:

WyriHaximus (approved):

PR #522, opened 2024-03-26
Drop leftover RingCentral PSR-7 dependency, use own PSR-7 implementation

clue:
This changeset drops the legacy RingCentral PSR-7 dependency and ensures we exclusively use our own PSR-7 implementation. This is the final step to replace the dated RingCentral implementation (#331) and allows us to eventually support PSR-7 v2 (#513). This is a purely internal change that comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply.

To re-iterate: Our own `Response`, `Request` and `ServerRequest` classes continue to work just like in previous versions. In particular, this builds on top of the recent PSR-7 changes for the `Response`, `Request`, `ServerRequest` and `Uri` classes (#518, #519, #520, #521). Unlike these PRs, this changeset doesn't show a noticeable impact on performance during my benchmarks. As a consequence, I consider this mostly an internal optimization (cleanup) only.

Given this changeset does not otherwise affect our public API, this should be safe to apply. The test suite confirms this has 100% code coverage and does not otherwise affect our APIs. If you want to explicitly install this dependency, you can still install it like this:

```bash
composer require ringcentral/psr7
```

If you enjoy this change and want to help us continue to ship more improvements, consider supporting this project, for example by [becoming a sponsor](https://github.com/sponsors/reactphp) ❤️

Builds on top of #521, #520, #519, #518, #480, #432, #431, #370, #170 and others

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,7 @@
         "react/event-loop": "^1.2",
         "react/promise": "^3 || ^2.3 || ^1.2.1",
         "react/socket": "^1.12",
-        "react/stream": "^1.2",
-        "ringcentral/psr7": "^1.2"
+        "react/stream": "^1.2"
     },
     "require-dev": {
         "clue/http-proxy-react": "^1.8",
--- a/examples/01-client-get-request.php
+++ b/examples/01-client-get-request.php
@@ -8,7 +8,7 @@ require __DIR__ . '/../vendor/autoload.php';
 $client = new Browser();
 
 $client->get('http://google.com/')->then(function (ResponseInterface $response) {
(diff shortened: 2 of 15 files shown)


SimonFrings (approved):
Went through all the changes, looks good to me :+1:

WyriHaximus (approved):
:tada: :tada: :tada: :tada: 

Seriously hoping everyone using this package doesn't rely on it for providing RingCentrals PSR-7 package. Not looking forward to the potential support tickets. We should put this in the release notes, and add it to the social media messages we'll put out

PR #523, opened 2024-03-27
Validate outgoing HTTP message headers and reject invalid messages

clue:
This changeset ensures we validate all outgoing HTTP message headers and reject any invalid messages. In particular, this ensures outgoing HTTP messages do not contain any headers with newlines.

Note that this does not affect normal operation and would only affect you if you're using invalid HTTP header names or values (such as when using untrusted user input). Likewise, this has no effect on incoming HTTP message headers which already use similar validation logic anyway (see #520 and others). This change comes with 100% code coverage and does not otherwise affect the public API, so it should be safe to apply.

I've originally planned to integrate this into our PSR-7 implementation (#518 and #519), but decided against this to not introduce any potential BC breaks. The suggested change still allows you to construct messages that would contain potentially invalid HTTP message fields, but would only reject such messages when trying to send over the wire.

### DIFF
--- a/src/Io/AbstractMessage.php
+++ b/src/Io/AbstractMessage.php
@@ -19,7 +19,7 @@ abstract class AbstractMessage implements MessageInterface
      * @internal
      * @var string
      */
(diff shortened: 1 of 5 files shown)


SimonFrings (approved):
Looks good, let's get this in :+1:

WyriHaximus (approved):
:ship: :it:

PR #528, opened 2024-05-24
Drop deprecated `Server` class, use `HttpServer` instead

clue:
This simple changeset drops the deprecated `Server` class from this package. As of `v1.5.0`, we recommend using the `HttpServer` class instead to avoid class name collisions, so upgrading should be straightforward for most use cases:

```php
// old (no longer supported)
$http = new React\Http\Server($handler);
$http = new React\Http\Server($middleware, $handler);
$http = new React\Http\Server($loop, $middleware, $handler);

// new (already supported since v1.5.0)
$http = new React\Http\HttpServer($middleware, $handler);
$http = new React\Http\HttpServer($loop, $middleware, $handler);
```

The original deprecation message has been in place for a couple of years already, so this should be relatively safe to apply. Additionally, a follow-up PR should drop the entire `$loop` argument as discussed in #517.

Builds on top of #417 and https://github.com/reactphp/socket/pull/314


SimonFrings (approved):
Looks good to me, let's get this shipped :+1:

WyriHaximus (approved):

PR #543, opened 2025-05-22
Run tests on PHP 8.4 and update test environment

PaulRotmann:
Builds on top of #536 and https://github.com/clue/framework-x/pull/267.

In order to avoid segfault with Xdebug 3.4.2 on PHP 8.1+ (https://bugs.xdebug.org/view.php?id=2332), i also updated the test suite to use PCOV instead of Xdebug on PHP 8+ as discussed in https://github.com/reactphp/socket/pull/323.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,11 @@ on:
 jobs:
   PHPUnit:
     name: PHPUnit (PHP ${{ matrix.php }})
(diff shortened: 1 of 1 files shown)


clue (approved):
@PaulRotmann Thanks for looking into this, changes LGTM! :shipit:

WyriHaximus (approved):

PR #544, opened 2025-05-26
Allow usage of environment variables in PHP ini settings that contain byte size values

graste:
I ran into a problem where the `IniUtil::iniSizeToBytes` function threw errors due to the usage of environment variables inside PHP ini settings to configure size values.

Example:

```ini
memory_limit=${PHP_MEMORY_LIMIT}
post_max_size=$PHP_POST_MAX_SIZE
```

This might be used e.g. in docker setups where the container gets `post_max_size`, `memory_limit` etc values via environment variables instead of providing different ini files per environments via mounts.

I suggest the changes within this pull request. I didn't add a changelog entry, as I don't know your rules for this. The used regular expression should work w/ PCRE and PCRE2 if I'm not mistaken. I didn't verify performance aspects, but usage of the function seems to be in constructors only. PS: That last string cast is not phpstan level 10 compatible. :dagger:

### DIFF
--- a/src/Io/IniUtil.php
+++ b/src/Io/IniUtil.php
@@ -19,6 +19,14 @@ final class IniUtil
             return (int)$size;
         }
 
+        // ini size values might be specified via environment variables, e.g.: memory_limit=${PHP_MEMORY_LIMIT}
+        $matches = [];
+        $envVarUsed = preg_match('#^\$(?|(\w+)|\{(\w+)})$#', $size, $matches);
+        $envVarName = $matches[1] ?? false;
+        if ($envVarUsed && $envVarName && array_key_exists($envVarName, $_ENV)) {
+            $size = (string)$_ENV[$envVarName];
+        }
+
         $suffix = \strtoupper(\substr($size, -1));
         $strippedSize = \substr($size, 0, -1);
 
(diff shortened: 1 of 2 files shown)


clue:
@graste Thanks for looking into this, and for linking the downstream issue too.

PHP's ini parser already interpolates `${VARNAME}` itself, checking ini directives first and falling back to the environment, so the value never reaches us as a literal:

```
$ MYSIZE=13M php -d 'post_max_size=${MYSIZE}' -r 'var_dump(ini_get("post_max_size"));'
string(3) "13M"
```

Same from PHP 5.6 through 8.5.

The bare `$MYSIZE` form is not valid ini syntax and is not interpolated, which is where a literal can survive:

```
$ MYSIZE=13M php -d 'post_max_size=$MYSIZE' -r 'var_dump(ini_get("post_max_size"));'
Warning: Invalid "post_max_size" setting. Invalid quantity "$MYSIZE": no valid leading digits, interpreting as "0" for backwards compatibility
string(7) "$MYSIZE"
```

The warning is PHP 8.2+, older versions accept it silently, but every version applies `0` here. Resolving it on our side would have us enforcing a limit PHP itself isn't.

Error reporting for an invalid value could certainly be better. An undefined variable interpolates to an empty string, and `iniSizeToBytes('')` then reports `" is not a valid ini size"` with the leading space, which is what phpstan/phpstan#5433 ran into. Worth improving, but the value is invalid to begin with, so that is its own change rather than env var support.

Closing here, but happy to accept PRs to improve error reporting 👍

